PR #2, opened 2022-08-29
feat: support file API reading

henryiii:
Uses the FileAPI to query information about the build system from CMake. Will likely be very important for PEP 660 support, extensionlib support, and Setuptools 62+ support for editable mode.

I originally used cattrs for the JSON reader, but dropped it due to https://github.com/python-attrs/cattrs/issues/253 - PyPy support was broken and was fixed quickly on April 15, but still hasn't been released. We can't afford to have PyPy broken by our dependencies, so no cattrs. But I still included the cattrs version for the tests. I loved the usage of exceptiongroup, so I kept that (and we use `__notes__` instead of `__note__`, awaiting merge for cattrs in https://github.com/python-attrs/cattrs/pull/303).

TODO: Finish the rest of the API, I was originally not covering everything when writing code model. Also consider if automatically loading jsonFile is best - one case (Directory) is odd because it looks like the class holds different info than the file it looks up.  I'm biased toward keeping this close to source, so maybe we could add the looked up file in an attribute instead.

@jcfr or @thewtex, any thoughts about supplying testing files for this?


henryiii:
This is entirely optional at the moment, so I think it's safe in, will likely change once we start needing it.

PR #3, opened 2022-09-07
refactor: use fspath for CMake

henryiii:
- chore: updates pre-commit checks
- refactor: use `__fspath__` for cmake path

### DIFF
--- a/tests/test_cmake_config.py
+++ b/tests/test_cmake_config.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import sys
 from pathlib import Path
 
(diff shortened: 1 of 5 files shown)


tests/test_cmake_config.py
```diff
@@ -9,10 +11,26 @@
 
 DIR = Path(__file__).parent.absolute()
 
+# Due to https://github.com/scikit-build/cmake-python-distributions/pull/279
```
henryiii, line 14: Not sure why this just hit this. Or why we didn't have to filter out `-GNinja` before.

PR #4, opened 2022-09-16
feat: adding config

henryiii:
Starting a config system. Using it looks like this:

```python
@dataclasses.dataclass
class SettingChecker:
    nested: NestedSettingChecker
    zero: Path
    one: str
    two: int
    three: List[str]
    four: List[int]
    five: str = "empty"
    six: Path = Path("empty")

@dataclasses.dataclass
class NestedSettingChecker:
    zero: Path
    one: str
    two: SettingChecker
    three: int = 3
```

That can be pulled from (currently) three sources:

```yaml
PREFIX_NESTED_ONE: "envvar"
```

```python
{"nested.one": "PEP 517 config"}
```

```toml
[tool.cmake]
nested.one = "TOML config"
```


henryiii:
I'm happy with the new configuration rewrite. One benefit over config-settings is that we are using completely vanilla dataclasses. That might become a bit of a pain if we are passing custom settings and through `typing.Annotated`, but for now this looks good. (I can't use config-settings because cattrs hasn't released since April with a fix for PyPy).

PR #24, opened 2022-10-19
feat: backport FindPython and include site-packages

henryiii:
* FindPython backport if CMake < 3.24
* site-packages is included via CMAKE_PREFIX_PATH
* Expanded mypy testing to tests (non-strict)

PR #25, opened 2022-10-20
fix: correct module issue and add tests

henryiii:
Pulling out changes from #24.

### DIFF
--- a/.github/codecov.yml
+++ b/.github/codecov.yml
@@ -1,3 +1,8 @@
 codecov:
   notify:
     after_n_builds: 10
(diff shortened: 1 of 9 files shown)


.github/codecov.yml
```diff
@@ -1,3 +1,8 @@
 codecov:
   notify:
     after_n_builds: 10
+  status:
+    project:
+      default:
+        target: auto
+        threshold: 5%
```
henryiii, line 8:
```suggestion
        threshold: 10%
```

PR #26, opened 2022-10-22
refactor: move get_requires into builder

henryiii:
Make a little more sense in here. Not top level.

### DIFF
--- a/src/scikit_build_core/build.py
+++ b/src/scikit_build_core/build.py
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from .get_requires import (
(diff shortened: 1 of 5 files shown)


henryiii:
Actually only get_requires belongs here.

src/scikit_build_core/build.py
```diff
@@ -2,7 +2,7 @@
 
 from __future__ import annotations
 
-from .get_requires import (
+from .builder.get_requires import (
     get_requires_for_build_wheel as skbuild_get_requires_for_build_wheel,
```
henryiii, line 6: This needs to be renamed

PR #33, opened 2022-10-29
fix: several fixes and a new simplest C test

henryiii:
- fix: exclude common directories
- fix: logging reports the correct file location on Python 3.8+
- fix: SKBUILD_PROJECT_NAME should be canonical
- tests: move test examples to project subdirectory
- tests: add test for simplist C-only like in README


henryiii:
@jcfr, @thewtex I hate writing things like 24ed940c07b8ce7796deb126496dac2300486e61 - at least it's supposed to be fixed in PyPy 7.3.10. Would it make sense to add a workaround in FindPython for PyPy < 7.3.9 though?

TL;DR: FindPython considers `SOABI` to be the source of truth, and it's been wrong in PyPy for years - it's missing the platform tag - while the other two related variables (`SO` & `EXT_SUFFIX`) are fine. PyPy will not load something without the platform tag, so you can't load an extension made from FindPython without renaming the extension.

https://foss.heptapod.net/pypy/pypy/-/issues/3816

henryiii:
It turns out this has been wrong so long it masked an issue with packaging, which means pip won't support correcting this value, so the change had to be reverted. Packaging can special case PyPy (just like it does for CPython), and then this won't matter, but for now, SOABI is unreliable and can't be used to compute the correct extension. I believe it either needs to be special cased or fixed in CMake as well. As it is, CMake's FindPython can't produce a binary that PyPy can load as an extension.

I wrote the workaround for pybind11 a year or two ago, I wish I'd pushed to get it fixed then.

henryiii:
Opened https://gitlab.kitware.com/cmake/cmake/-/issues/24121 (turns out I'd opened one two years ago too, and a few other people have opened one since).

PR #34, opened 2022-10-29
feat: automatic versioning

henryiii:
If we did want to support VCS versioning, this is what it would look like. Not
sure it's worth it - the version is already single-sourced, and this requires a
bit more work to package. But, then again, not being able to forget to change
the version is really nice.


henryiii:
We might want to be moving along fairly fast, so let's do this for now. We can always reduce the dependency later.

PR #40, opened 2022-11-04
feat: prettier logging with config setting

henryiii:
Not supported in setuptools mode. Will take some thought there.

To test this or similar things, you can run:

```bash
pipx run build --wheel
PIP_FIND_LINKS="dist" pipx run build --wheel tests/packages/simple_pyproject_ext -o dist -C scikit-build.log-level=NOTSET
```

### DIFF
--- a/src/scikit_build_core/pyproject/wheel.py
+++ b/src/scikit_build_core/pyproject/wheel.py
@@ -13,6 +13,7 @@ from ..builder.builder import Builder
 from ..builder.wheel_tag import WheelTag
 from ..cmake import CMake, CMakeConfig
 from ..settings.skbuild_settings import read_settings
+from .init import setup_logging
 
 __all__: list[str] = ["build_wheel"]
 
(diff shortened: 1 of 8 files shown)


src/scikit_build_core/pyproject/wheel.py
```diff
@@ -94,8 +94,13 @@ def build_wheel(
...
 
-        out = wheel.build({"platlib": str(install_dir)}, tags=tags.tags_dict())
-        shutil.move(out, wheel_directory)
+        out = Path(wheel.build({"platlib": str(install_dir)}, tags=tags.tags_dict()))
+        wheel_path = Path(wheel_directory) / out.name
+        try:
+            out.rename(wheel_path)
+        except FileExistsError:
+            wheel_path.unlink()
+            out.rename(wheel_path)
```
henryiii, line 103: This is a fix, wheels should overwrite

tests/test_module_dir.py
```diff
@@ -31,6 +31,7 @@ def test_all_modules_filter_all():
     all_modules = on_all_modules("scikit_build_core", pkg=False)
     all_modules = (n for n in all_modules if not n.split(".")[-1].startswith("_"))
     for name in all_modules:
+        print(name)
```
henryiii, line 34: Was debugging, probably can remove

henryiii:
@mayeut Something’s wrong with ninja on GHA. Just restarting this causes it to fail when it used to pass. I’ll try to look at it in the next day or two. We are supposed to be protecting against this, but if the ninja package itself was broken, that might cause this.

mayeut:
Just saw https://github.com/scikit-build/ninja-python-distributions/issues/157

PR #43, opened 2022-11-06
fix: handle `PermissionError` in reading `libdir.is_dir()`

agoose77:
`pyodide-build` fails with `scikit-build-core`, because `scikit-build-core` tries to `stat()` the `libdir` given by `sysconfig`. I'm not sure what the bug is here; should we be able to stat `LIBDIR` under `pyodide-build`? I suspect not if we're pretending to be the target environment, but I'm not certain. 

This is a very specific fix, and I suspect there's a nicer way of generalising this to the  class of "this isn't a conventional Python distribution" errors. However, this patch as-it-stands fixes `pyodide-build`, so it's a starting point for a proper fix.

### DIFF
--- a/src/scikit_build_core/builder/sysconfig.py
+++ b/src/scikit_build_core/builder/sysconfig.py
@@ -21,22 +21,21 @@ def get_python_library() -> Path | None:
     multiarch = sysconfig.get_config_var("MULTIARCH")
     masd = sysconfig.get_config_var("multiarchsubdir")
 
-    try:
-        libdir_is_dir = libdir and ldlibrary and libdir.is_dir()
-    except PermissionError:
-        return None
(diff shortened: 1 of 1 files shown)


henryiii:
Getting the lib for is extremely unimportant - it only matters at all for Windows, and I believe we’re still ignoring it for FindPython.

henryiii:
What happens if exists is used instead of is_dir? Does it just move the permission error down? (Also would want it to return None for permission errors, but just wondering)

Also there might be a better way to handle cross compiles.

henryiii:
> next step from your comment is

I was just says this was fine, finding the lib is not very important.

henryiii:
Thanks!

PR #47, opened 2022-11-07
feat: support for setting python & abi tag (including limited API)

### LINKED ISSUE #39: Support for Limited ABI / Stable API
We should think about how to support the Limited ABI / Stable API. This requires the following:

* Wheels need to be set with specific `abi3` tag and the minimum cp version.
* Code should be compiled with the correct `Py_LIMITED_API` define.
* On Windows, "python3.dll" must be used instead of "python39.dll" for linking. I'm not sure FindPython supports this! (Edit: it doesn't, opened request)
* A limited ABI extension has a special EXT_SUFFIX, not sure how to grab that, and not sure FindPython supports this! (Edit: it doesn't, but we have to require users overwrite this to support PyPy anyway)

This might need to integrate nicely with pythoness wheels, which we need for cmake-python-distributions and such.

In case I don't commit it for a while, here's the code for playing with tags at the bottom of `wheel_tag.py`:

```python
if __name__ == "__main__":
    import argparse

    parser = argparse.ArgumentParser()
    parser.add_argument(
        "--archs",
        nargs="*",
        default=[],
        help="Specify one or more archs (macOS only currently)",
    )
    parser.add_argument(
        "--abi", help="Specify abi min version, like 'cp37'. Use 'py3' to be pythonless"
    )
    args = parser.parse_args()
    tag = WheelTag.compute_best(args.archs)
    if args.abi:
        tag.abis = ["none" if args.abi.startswith("py") else "abi3"]
        tag.pyvers = args.abi.split(".")
    print(tag)
```

This would assume a setting of "cp38" (abi3) or "py2.py3" (no Python).

Ahh, due to optional compiled extensions, we'll have to support pure Python tags & some controllable way to indicate that a binary can be skipped. (Probably expose a hook, maybe make some cases statically configurable). Or we could inspect the wheel & decide based on CMake's output, and handle it entirely from there.

henryiii:
Supported in both modes. Should also support Pythonless wheels. Upstream request at https://gitlab.kitware.com/cmake/cmake/-/issues/24141. Closes #39.

Setuptools support is pretty bad due to wheel not having a public API, so in setuptools mode this should still be specified in the same way (pyproject.toml), rather than by setting `py_limited_abi`, so you get the proper wheel name. But it should still work as before, you just have to also set up the wheel name if you only specify `py_limited_api`.

### DIFF
--- a/src/scikit_build_core/builder/wheel_tag.py
+++ b/src/scikit_build_core/builder/wheel_tag.py
@@ -28,7 +28,7 @@ class WheelTag:
     # TODO: plats only used on macOS
     @classmethod
     def compute_best(
-        cls: type[Self], archs: Sequence[str] = (), abi_tag: str = ""
+        cls: type[Self], archs: Sequence[str] = (), py_abi_tag: str = ""
     ) -> Self:
         best_tag = next(packaging.tags.sys_tags())
         interp, abi, *plats = (best_tag.interpreter, best_tag.abi, best_tag.platform)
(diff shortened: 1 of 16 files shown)


src/scikit_build_core/builder/wheel_tag.py
```diff
@@ -67,3 +74,23 @@ def tags_dict(self) -> dict[str, list[str]]:
...
+    parser.add_argument(
+        "--archs",
+        nargs="*",
+        default=[],
+        help="Specify one or more archs (macOS only currently)",
+    )
+    parser.add_argument(
+        "--abi",
+        default="",
+        help="Specify abi min version, like 'cp37'. Use 'py3' to be pythonless",
```
henryiii, line 92:
```suggestion
        help="Specify py-abi, like 'cp37-abi3'. Use 'py3-none' to be pythonless",
```

Issue #62, opened 2022-11-09
Reproducible builds


henryiii:
See https://hatch.pypa.io/latest/plugins/utilities/#hatchling.builders.utils.get_reproducible_timestamp and such.

We should set the date to be the latest time for the source files.

PR #64, opened 2022-11-09
feat: support reproducible sdist builds

agoose77:
Hey @henryiii,

This is rudimentary, but sets the scene for reproducible builds. I don't love the ergonomics and you seem to have a design vision here, so feel free to rewrite / ignore this if it needs another pass.

I'm not sure whether we need to offer an option for non-reproducible builds. I don't know enough about when you *wouldn't* want this to make a strong decision. But, it might also be something that the Hatch builder can deal with down the line, and sciki-build-core's backend can just be opinionated.

### DIFF
--- a/src/scikit_build_core/pyproject/sdist.py
+++ b/src/scikit_build_core/pyproject/sdist.py
@@ -40,7 +40,7 @@ def normalize_file_permissions(st_mode: int) -> int:
     permissions can vary on systems with different umasks. Normalising
     to 644 (non executable) or 755 (executable) makes builds more reproducible.
 
-    Taken from https://github.com/takluyver/flit/blob/6a2a8c6462e49f584941c667b70a6f48a7b3f9ab/flit_core/flit_core/common.py#L257
+    Taken from https://github.com/pypa/flit/blob/6a2a8c6462e49f584941c667b70a6f48a7b3f9ab/flit_core/flit_core/common.py#L257
     """
     # Set 644 permissions, leaving higher bits of st_mode unchanged
     new_mode = (st_mode | 0o644) & ~0o133
--- a/src/scikit_build_core/pyproject/sdist.py
+++ b/src/scikit_build_core/pyproject/sdist.py
@@ -1,7 +1,12 @@
 from __future__ import annotations
 
+import contextlib
+import copy
+import gzip
 import io
+import os
 import tarfile
+import time
 from pathlib import Path
 
 from pyproject_metadata import StandardMetadata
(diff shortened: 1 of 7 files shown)


src/scikit_build_core/pyproject/sdist.py
```diff
@@ -18,6 +23,49 @@ def __dir__() -> list[str]:
...
+
+
+def normalize_file_permissions(st_mode: int) -> int:
+    """
+    Normalize the permission bits in the st_mode field from stat to 644/755
+    Popular VCSs only track whether a file is executable or not. The exact
+    permissions can vary on systems with different umasks. Normalising
+    to 644 (non executable) or 755 (executable) makes builds more reproducible.
+
+    Taken from https://github.com/takluyver/flit/blob/6a2a8c6462e49f584941c667b70a6f48a7b3f9ab/flit_core/flit_core/common.py#L257
```
henryiii, line 43:
```suggestion
    Taken from https://github.com/pypa/flit/blob/6a2a8c6462e49f584941c667b70a6f48a7b3f9ab/flit_core/flit_core/common.py#L257
```

Let's not repeat hatch by copying the (now) wrong path to flit. :)

henryiii:
X-ref: first half of #62.

src/scikit_build_core/pyproject/sdist.py
```diff
@@ -31,22 +79,38 @@ def build_sdist(
     with Path("pyproject.toml").open("rb") as f:
         pyproject = tomllib.load(f)
 
+    if settings.sdist.reproducible:
+        transform_tar_info = normalize_tar_info
+        timestamp = get_reproducible_epoch()
+    else:
+
+        def transform_tar_info(tar_info):
+            return tar_info
```
henryiii, line 88: This bothers me - why is it passing the type check?

henryiii: Ah, I think I see why. That's embarrassing...

henryiii:
What do you not like about the ergonomics, anything concrete?

src/scikit_build_core/pyproject/sdist.py
henryiii: Yep, that's why. Fixed.

henryiii:
>  I think I'm not in-love about the tarfile API

I really just wish tarfile and zipfile would be more similar. I do really like `zipfile.Path` (3.8+), but it didn't make it into tarfile. `¯\_(ツ)_/¯`

> What are your thoughts on exposing the reproducibility setting? 

That's a good question. I think it's okay to have it, it's pretty low cost and at the least it might help expose that the builds are reproducible by default. :) But I'm also always okay to wait until someone needs something to add config.

henryiii:
I'd not worry too much about sorting, we can do it in a followup if it matters. I wonder if we could store the SHA from a run and then verify that a build always produces that? That would avoid a 2 second delay and would work across platforms - seeing Windows produce the same file SHA as Linux would be reassuring for reproducibility!

henryiii:
It seems like this is reproducible on Python 3.9+ on Unix?

henryiii:
Have you thought about / looked into wheels too? :)

Also, one of us should compare <3.9 and >=3.9 results to see what changed eventually.

henryiii:
Looks like it's this:

https://github.com/python/cpython/commit/674935b8caf33e47c78f1b8e197b1b77a04992d2
(from https://github.com/python/cpython/issues/63019). Not a problem, just dual values.

So we just need to solve the Windows issue now. :)

So there will be a before 3.9 and after 3.9 value.

PR #126, opened 2022-11-21
fix: paths on Windows

henryiii:
- ci: MSVC 2019 test (2022 requires CMake 3.21+)
- fix: an issue with setting paths in the CLI on Windows

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
       fail-fast: false
       matrix:
         python-version: ["3.7", "pypy-3.8", "3.11"]
-        runs-on: [ubuntu-latest, macos-latest, windows-latest]
+        runs-on: [ubuntu-latest, macos-latest, windows-2022]
 
         include:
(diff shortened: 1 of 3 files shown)


tests/test_pyproject_pep518.py
```diff
@@ -113,7 +121,9 @@ def test_pep518_wheel(virtualenv, build_args):
 @pytest.mark.integration
 @pytest.mark.isolated
 def test_pep518_pip(virtualenv):
-    virtualenv.run(f"python -m pip install -v {HELLO_PEP518}")
+    virtualenv.run(
+        f"python -m pip install --config-setting=logging.level=INFO -v {HELLO_PEP518}"
```
henryiii, line 125:
```suggestion
        f"python -m pip install --config-settings=logging.level=INFO -v {HELLO_PEP518}"
```

henryiii:
`pypy-3.8 on windows-2022` doesn't support build running inside a venv, PyPy on Windows seems to not like venv in venv?

PR #161, opened 2022-12-14
fix: `entrypoints.txt` should be `entry_points.txt`

njzjz:
This fixes the `entry_points`, including `[project.scripts]`.


henryiii:
🤦

Thanks!

PR #162, opened 2022-12-14
fix: Windows ARM cross-compile

henryiii:
Experimental cross-compile support.

### DIFF
--- a/src/scikit_build_core/builder/sysconfig.py
+++ b/src/scikit_build_core/builder/sysconfig.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import configparser
 import os
 import sys
 import sysconfig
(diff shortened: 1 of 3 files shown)


src/scikit_build_core/builder/sysconfig.py
```diff
@@ -30,7 +31,18 @@ def __dir__() -> list[str]:
     return __all__
 
 
-def get_python_library() -> Path | None:
+def get_python_library(env: Mapping[str, str], *, abi3: bool = False) -> Path | None:
+    # When cross-compiling, check DIST_EXTRA_CONFIG first
+    config_file = env.get("DIST_EXTRA_CONFIG", None)
+    if config_file and Path(config_file).is_file():
```
henryiii, line 37:
```suggestion
    if sys.platform.startswith("win") and config_file and Path(config_file).is_file():
```

PR #175, opened 2023-01-20
chore: use ruff

henryiii:
Trying this out. Very impressed so far. Some minor issues, but very complete, fast, and simple. No more `.flake8` file!

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -53,7 +53,7 @@ repos:
         additional_dependencies: [black==22.8.0]
 
   - repo: https://github.com/charliermarsh/ruff-pre-commit
-    rev: v0.0.226
+    rev: v0.0.228
     hooks:
       - id: ruff
         args: ["--fix"]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -24,10 +24,6 @@ repos:
   - repo: https://github.com/pre-commit/pygrep-hooks
     rev: v1.10.0
     hooks:
-      - id: python-check-blanket-noqa
-      - id: python-check-blanket-type-ignore
-      - id: python-no-eval
-      - id: python-use-type-annotations
       - id: rst-backticks
       - id: rst-directive-colons
       - id: rst-inline-touching-normal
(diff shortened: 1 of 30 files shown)


noxfile.py
```diff
@@ -61,7 +61,6 @@ def docs(session: nox.Session) -> None:
 
     if session.posargs:
         if "--serve" in session.posargs:
-            print("Launching docs at http://localhost:8000/ - use Ctrl-C to quit")
```
henryiii, line 64: Oh, wow, auto fixing no print statements might be something to turn off. :)

pyproject.toml
```diff
@@ -94,14 +94,12 @@ cmake_source_dir = "scikit_build_core.setuptools.extension:cmake_source_dir"
 build.exclude = ["extern"]
 version.source = "vcs"
 build.hooks.vcs.version-file = "src/scikit_build_core/_version.py"
-
-[tool.hatch.build.force-include]
```
henryiii, line 98: This is a bug in the TOML crate for Rust, see https://github.com/charliermarsh/ruff/issues/1894.

noxfile.py
agoose77: Huh that seems a bit aggressive!

.pre-commit-config.yaml
```diff
@@ -56,39 +52,11 @@ repos:
...
-          - flake8-print
-
-  - repo: https://github.com/pycqa/flake8
-    rev: 6.0.0
-    hooks:
-      - id: flake8
-        exclude: docs/conf.py
-        additional_dependencies: *flake8_dependencies
+      - id: ruff
+        args: ["--fix"]
```
henryiii, line 59: This really should have --force-exclude, but the next version of the Action won't need it. https://github.com/charliermarsh/ruff-pre-commit/pull/20

.pre-commit-config.yaml
```diff
@@ -56,39 +52,11 @@ repos:
       - id: blacken-docs
         additional_dependencies: [black==22.8.0]
 
-  - repo: https://github.com/PyCQA/isort
-    rev: 5.11.4
+  - repo: https://github.com/charliermarsh/ruff-pre-commit
+    rev: v0.0.226
```
henryiii, line 56:
```suggestion
    rev: v0.0.228
```

.pre-commit-config.yaml
henryiii: And it's out!

henryiii:
pre-commit.ci autofix

PR #181, opened 2023-01-27
feat: build directory setting & reuse

henryiii:
Support for custom build directory, opt-in. Using `FORCE` for the moment. Doesn't try to rewrite the paths for moving CMake yet, but the info for this is there in the new json file.

Maybe this shouldn't have a `cmake.` prefix? Edit: yes, looks like we could sync-up with meson-python if we just use `build-dir`, so let's go with that. See https://github.com/mesonbuild/meson-python/issues/275 for `builddir` -> `build-dir` there.

- [x] TODO: this should be able to (maybe via substitution?) have a unique path per interpreter.

Might adjust this later. My thought is we can make this opt-in for now, then if we make it opt-out, we can protect the change on a minimum version selection.

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,6 +179,9 @@ strict-config = true
 # scikit-build-core version.
 minimum-version = "0.1"  # current version
 
+# Build directory (empty will use a temporary directory)
+build-dir = ""
+
 [tool.scikit-build.cmake.define]
 # Put CMake defines in this table.
 ```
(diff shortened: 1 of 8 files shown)


henryiii:
This still doesn't provide any benefit, since the Python directories change between invocations. We could either sed one for the other, or we could declare that only non-isolated builds can benefit (at least the changing paths isn't breaking anything).

PR #184, opened 2023-01-28
fix: remove duplicate ruff code

burgholzer:
This tiny PR removes a duplicate code definition in the `ruff` configuration.


henryiii:
Thanks! I noticed this briefly after sorting but then forgot about it.

PR #231, opened 2023-03-17
fix: version not a string (typing updates)

henryiii:
Fixes to upstream were released quicker than I thought! :)


src/scikit_build_core/setuptools/extension.py
```diff
@@ -94,7 +97,7 @@ def build_extension(self, ext: setuptools.Extension) -> None:
         builder.configure(
             defines=defines,
             name=dist.get_name(),
-            version=dist.get_version(),
+            version=Version(dist.get_version()),
```
henryiii, line 100: This was an actual mistake! :D

PR #251, opened 2023-03-30
feat: rework plugins

### LINKED ISSUE #236: Local dynamic metadata plugins
We should allow dynamic metadata plugins to be local to the package. This could either be done by using a path instead of a module name, or by adding a metatdata-path(s?) setting. Thoughts, @bennyrowland?

henryiii:
Closer to current proposal, closes #236.

New features:

* Uses `provider =`
* New `provider-path =` for local plugins
* Custom or local plugins require `tool.scikit-build.experimental=true` to use, since API is not stable yet.
* `Dict[str, Any]` support for TOML reader.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -372,8 +372,11 @@ CMAKE_ARGS: -DSOME_DEFINE=ON;-DOTHER=OFF
 
 ## Dynamic metadata (WIP)
 
-Scikit-build-core 0.3.0 will support dynamic metadata. This is experimental
-until a release. The public interface for making a plugin is very experimental.
+Scikit-build-core 0.3.0 will support dynamic metadata. This is not ready for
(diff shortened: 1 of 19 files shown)


bennyrowland:
@henryiii - nice work, this is all evolving rather quickly now. Sorry for being out of touch for a little bit, I have been in bed with COVID for the last week and not up to doing much, but I am still actively engaged with this project and hope to continue to contribute.

henryiii:
That's okay, focus on getting better! Hopefully this is something that can be released (not immediately!), and then we can see over time what people are happy with as we work on (3!) PEPs. I'll make sure you are a collaborator once I set up the repo(s) for those. I'll probably aim to bring this up at the packaging summit in a few weeks at PyCon.

I'll probably focus on editable installs next, so we can be ready for a release.

PR #254, opened 2023-04-03
fix: support script rewriting

### LINKED ISSUE #130: Support scripts properly
See https://packaging.python.org/en/latest/specifications/binary-distribution-format/#recommended-installer-features & https://github.com/mesonbuild/meson-python/issues/220. We should process scripts in `/scripts` to replace a nice shebang with the required one.

This is what setuptools uses (and misses pypy):

```python
shebang_pattern = re.compile('^#!.*python[0-9.]*([ \t].*)?$')
```

Also seems to not suppert `pythonw`. Though I'm not sure you'd indicate that in a source file that you wanted to run on Linux too.

henryiii:
Closes #130. Properly processes and rewrites script's shebang lines.

### DIFF
--- a/src/scikit_build_core/build/_scripts.py
+++ b/src/scikit_build_core/build/_scripts.py
@@ -1,0 +1,31 @@
+from __future__ import annotations
+
+import re
+from pathlib import Path
+
+__all__ = ["process_script_dir"]
+
+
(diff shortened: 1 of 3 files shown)


henryiii:
NumPy is building from source (9 mins).

PR #256, opened 2023-04-04
tests: fix msys failures

henryiii:
Testing to see if I can fix the `address space needed by '_common.cpython-311-x86_64-msys.dll' (0x620000) is already occupied` failures. Switching to UCRT instead, currently disabling a few setuptools tests to get it to pass. Speeding up the tests, too.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,7 +131,7 @@ jobs:
         run: python3.9 -m pytest -ra --showlocals -m "not virtualenv"
 
   msys:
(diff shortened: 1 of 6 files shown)


.github/workflows/ci.yml
```diff
@@ -141,13 +141,13 @@ jobs:
     steps:
       - uses: msys2/setup-msys2@v2
         with:
-          msystem: msys
+          msystem: ucrt-x86_64
```
henryiii, line 144:
```suggestion
          msystem: UCRT64
```

PR #260, opened 2023-04-05
tests: use fixtures for packages

henryiii:
Simpler testing.


henryiii:
Ahh, pytest is exceeding the maximum path length on Windows. 😡

henryiii:
We are _very_ close to the limit - a long test name would put it over (pytest test caps the test name to dir name at 30 chars). But using `pkg` for the inner folder name does keep it under the limit for now.

PR #270, opened 2023-04-12
tests: adjust coverage, use --cov in nox

henryiii:
Also printing some durations.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -52,7 +52,7 @@ def tests(session: nox.Session) -> None:
 
     install_arg = "-e.[test,cov]" if "--cov" in posargs else "-e.[test]"
     session.install(install_arg, *extra)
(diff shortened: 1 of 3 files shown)


noxfile.py
```diff
@@ -49,20 +50,11 @@ def tests(session: nox.Session) -> None:
     if (3, 8) <= sys.version_info < (3, 12):
         extra.append("numpy")
 
-    session.install("-e.[test]", *extra)
+    install_arg = "-e.[test,cov]" if "--cov" in posargs else "-e.[test]"
+    session.install(install_arg, *extra)
     session.run("pytest", *session.posargs, env=env)
```
henryiii, line 55:
```suggestion
    session.run("pytest", *posargs, env=env)
```

PR #279, opened 2023-04-14
typo on tab Fortran (was Cython)

zerothi:
A typo in the documentation.

The tab was Cython | Cython, but the latter should have been fortran.


henryiii:
Thanks for the help on the docs! They are quite new and not fully polished yet, this has been helpful!

henryiii (approved):

PR #323, opened 2023-05-18
feat: source-dir for pyproject.toml

henryiii:
Adding source-dir.


tests/test_skbuild_settings.py
```diff
@@ -192,6 +197,7 @@ def test_skbuild_settings_pyproject_toml(tmp_path, monkeypatch):
             cmake.define = {a = "1", b = "2"}
             cmake.build-type = "Debug"
             cmake.verbose = true
+            cmake.source_dir = "a/b/c"
```
henryiii, line 200:
```suggestion
            cmake.source-dir = "a/b/c"
```

Issue #335, opened 2023-05-22
CMAKE_ARGS delimiter


eirrgang:
https://scikit-build-core.readthedocs.io/en/latest/configuration.html#configuring-cmake-arguments-and-defines shows a semi-colon (`;`) delimiter in the CMAKE_ARGS list, but the delimiter is not explicitly mentioned.

CMAKE_ARGS is mentioned at the bottom of the section https://scikit-build-core.readthedocs.io/en/latest/index.html#configuration and it presumably uses the semicolon delimiter mentioned in the sample config above, where a comment describes how to override the yaml array from the command line.

However, the CMAKE_ARGS may be treated differently when passed as an environment variable. The retrieval code (currently in `Builder.get_cmake_args()`) has always been `self.config.env.get("CMAKE_ARGS", "").split(" ")`, it seems. There may be some place where tokens are further split by `;`, but I'm not seeing it.

eirrgang:
It looks like `split(";")` happens in `BuildCMake.finalize_options()` for `BuildCMake.cmake_args`. It isn't clear to me whether this applies to CMAKE_ARGS from the environment at an appropriate time.

eirrgang:
#339 (and comments at #336) clarify that `CMAKE_ARGS` is intended to be space-delimited.

#339 updates the line that begins "Beyond the normal `SKBUILD_CMAKE_ARGS`", but I can't find any other mention of SKBUILD_CMAKE_ARGS. Maybe something was accidentally elided in a previous merge?

PR #336, opened 2023-05-22
Handle " " or ";" delimiter for CMAKE_ARGS env var.

eirrgang:
The documentation implies that ";" is the correct delimiter. This change ensures that a CMAKE_ARGS environment variable string will be split on ";" when it is read. Split on " " is preserved, as well.

Ref #335.

### DIFF
--- a/src/scikit_build_core/builder/builder.py
+++ b/src/scikit_build_core/builder/builder.py
@@ -71,7 +71,11 @@ class Builder:
         """
         # Adding CMake arguments set as environment variable
         # (needed e.g. to build for ARM OSX on conda-forge)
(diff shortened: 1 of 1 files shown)


eirrgang:
The ";" delimiter seemed to work previously. It is not clear to me from documentation whether CMAKE_ARGS is intended to be space delimited or ";". If it is intended to be space delimited, this MR could be discarded in favor of a documentation tweak.

henryiii:
The main change in 0.4 is that Ninja is selected more consistently if it's available. That seems to affect some processing for some reason.

I think the idea was that `SKBUILD_CMAKE_ARGS` was matches the rest of the arguments and uses `;`, while `CMAKE_ARGS` is a compatibility input (largely for conda-forge) that uses spaces (because we don't have control over that).

eirrgang:
superseded by #339

PR #352, opened 2023-06-02
fix: abi3 backwards & Python_SABI_LIBRARY bug

henryiii:
Discovery doesn't work when cross compiling, so it's important these are set.

Also need to set Python_SABI_LIBRARY separately.

### DIFF
--- a/src/scikit_build_core/builder/builder.py
+++ b/src/scikit_build_core/builder/builder.py
@@ -134,7 +134,9 @@ class Builder:
                 limited_abi = False
 
         python_library = get_python_library(self.config.env, abi3=False)
-        python_sabi_library = get_python_library(self.config.env, abi3=False) if limited_abi else None
+        python_sabi_library = (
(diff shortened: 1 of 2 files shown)


src/scikit_build_core/builder/builder.py
```diff
@@ -129,11 +129,14 @@ def configure(
...
                 target_minor_version = int(self.settings.wheel.py_api[3:])
-                limited_abi = target_minor_version >= sys.version_info.minor
+                limited_abi = target_minor_version <= sys.version_info.minor
             else:
                 limited_abi = False
 
-        python_library = get_python_library(self.config.env, abi3=limited_abi)
+        python_library = get_python_library(self.config.env, abi3=False)
+        python_sabi_library = (
+            get_python_library(self.config.env, abi3=False) if limited_abi else None
```
henryiii, line 138:
```suggestion
            get_python_library(self.config.env, abi3=True) if limited_abi else None
```

PR #396, opened 2023-06-21
fix: backport CMake 3.27's FindPython

henryiii:
This might fix the Windows ARM issues.


henryiii:
Will wait until there's something required to update the backport version.

PR #414, opened 2023-06-24
fix: avoid interfering with other setuptools plugins

### LINKED ISSUE #413: setuptools patching (semi-randomly) breaks setuptools-rust, resulting in nonfunctional packages
If `setuptools-rust` and `scikit-build-core` are installed within the same build environment, depending on plugin load order, the latter could cause Rust extensions to be silently skipped and nonfunctional packages to be built.

The problem is that both packages define `setuptools.finalize_distribution_options` entry points, and both override `has_ext_modules` method:

- https://github.com/scikit-build/scikit-build-core/blob/9ed353902d2ee3af33f70025c8e46ec075e0b834/src/scikit_build_core/setuptools/build_cmake.py#L185
- https://github.com/PyO3/setuptools-rust/blob/0ca3eb47bbadd7b4de32c302bf0380500187a807/setuptools_rust/setuptools_ext.py#L277-L279

If the entry point from setuptools-rust is executed first, then scikit-build-core overwrites its `has_ext_modules` method. As a result, the Rust extensions aren't detected, `build` command doesn't invoke `build_rust` and the package is installed without any Rust extensions.

FWICS setuptools-rust avoids the problem by storing and calling the previous value of `has_ext_modules` rather than the generic method from Distribution. FWICS doing the same in scikit-build-core fixes building Rust extensions (but I haven't tested whether it doesn't break something related to scikit-build-core). I tested changing the linked line to:

```
    orig_has_ext_modules = dist.has_ext_modules
    dist.has_ext_modules = lambda: orig_has_ext_modules() or _has_cmake(dist)
```

Gentoo bug report: https://bugs.gentoo.org/909035

henryiii:
Fix #413.

### DIFF
--- a/src/scikit_build_core/setuptools/build_cmake.py
+++ b/src/scikit_build_core/setuptools/build_cmake.py
@@ -184,7 +184,7 @@ def _prepare_extension_detection(dist: Distribution) -> None:
 
     orig_has_ext_modules = dist.has_ext_modules
     dist.has_ext_modules = lambda: orig_has_ext_modules() or _has_cmake(dist)  # type: ignore[method-assign]
-    
+
     # Hack for stdlib distutils
     if not setuptools.distutils.__package__.startswith("setuptools"):  # type: ignore[attr-defined]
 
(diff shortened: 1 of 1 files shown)


mgorny:
Thanks!

PR #422, opened 2023-07-05
fix: Add purelib site-packages to sys.path

hesiod:
In order to actually discover entry points provided by dependencies, the purelib site-packages need to be part of sys.path when looking for entry points. This is not the case by default when building with PEP517 build isolation enabled.

I've found that without this PR, there is no way to actually utilize `cmake.prefix` or `cmake.module` since the entry points never get picked up by scikit-build-core.


hesiod:
While trying to narrow down a simple test case for the issue, I realized that the entry points weren't visible due to some accidental misconfiguration on my end - so it turns out I was chasing a red herring with the `sys.path` idea. Sorry for the noise!

henryiii:
No problem, love that someone's trying these out! I hope to use them for some things soon, but so far I haven't had a chance to dogfood this feature. :)

PR #423, opened 2023-07-07
chore: use get_origin/get_args

henryiii:
Simpler & nicer.


henryiii:
Don't think the fedora issue is mine:

```
Error: 
 Problem: conflicting requests
  - nothing provides python(abi) = 3.11 needed by python3-cattrs-22.2.0-13.fc39.noarch from fedora
  - nothing provides python3.11dist(attrs) >= 20 needed by python3-cattrs-22.2.0-13.fc39.noarch from fedora
```

Issue #437, opened 2023-07-18
Presence of hidden files breaks editable installs


vyasr:
Git supports nested `.gitignore` files. The presence of such a file inside the project's source currently breaks editable installs with the following traceback:

```
...                                                                                                                                           
        File "scikit_build_core/build/__init__.py", line 54, in build_editable
          return _build_wheel_impl(
        File "scikit_build_core/build/wheel.py", line 254, in _build_wheel_impl
          modules = {
        File "scikit_build_core/build/wheel.py", line 255, in <dictcomp>
          path_to_module(Path(v).relative_to(wheel_dirs["platlib"])): str(
        File "scikit_build_core/build/_pathutil.py", line 26, in path_to_module
          path = path.with_name(path.name.split(".", 1)[0])
        File "/lib/python3.10/pathlib.py", line 762, in with_name
          raise ValueError("Invalid name %r" % (name))
      ValueError: Invalid name ''
      [end of output]
```

The problem is [here](https://github.com/scikit-build/scikit-build-core/blob/main/src/scikit_build_core/build/_pathutil.py#L26) because `".gitignore".split(".", 1)[0] == ""`, which is an invalid argument to `path.with_name`.

We could fix this specifically for `.gitignore` by adding it to [the list of ignored files](https://github.com/scikit-build/scikit-build-core/blob/main/src/scikit_build_core/build/_file_processor.py#L12). However, that wouldn't fix the general case of adding an arbitrary hidden file (e.g. consider config files for linters like .flake8 or tools like .bumpversion). The current easiest workaround would be to add such files to `sdist.exclude`, but that wouldn't work if a hidden file was actually important for the functioning of a package (I'm less concerned with linter configs etc since nobody should be developing off an sdist).

vyasr:
#444 says it's only a partial solution, did the other bits get completed too somewhere?

henryiii:
Two more things probably should be done: revisiting the default ignore, and we maybe should support nested gitignores. I guess I can keep this open, though one/two separate issues might be better.

vyasr:
Feel free to track however you want, I just asked for clarification.

henryiii:
:robot: _AI text below_ :robot:

This is no longer an issue and can be closed.

The crash was that a hidden file (e.g. `.gitignore`, `.flake8`, `.clang-tidy`) reached `path_to_module`, where `".gitignore".split(".", 1)[0] == ""` produced an invalid argument to `Path.with_name`.

[...]

PR #438, opened 2023-07-18
Add gitignore to list of ignored files

vyasr:
I didn't add a test, but would be happy to do so. The reason I waited is that it seems like the easiest option would be to add the file to one of the existing tests/packages examples, but that would require modifying the hashes and I wasn't sure if that would be the preferred result.

Partially addresses #437, but only for the special case of gitignore. This doesn't fix the more general case of hidden files.

### DIFF
--- a/src/scikit_build_core/build/_file_processor.py
+++ b/src/scikit_build_core/build/_file_processor.py
@@ -11,6 +11,7 @@ __all__: list[str] = ["each_unignored_file"]
 
 EXCLUDE_LINES = [
     ".git/",
+    ".gitignore",
     ".tox/",
     ".nox/",
     ".egg-info/",


henryiii:
I think it's important to keep the .gitignore(s) in the SDist - it should only excluded from wheels. Maybe we should have some better defaults, and some control over the defaults. I've got an idea, and will try to get back on this in a bit (in the middle of four conferences in three weeks!).

The key problem in https://github.com/scikit-build/scikit-build-core/issues/437 would be fixed by fixing the logic for files that start with a dot, which are not valid packages so should be skipped.

henryiii:
Does the current version (0.4.8) work, by the way? It should be "fixed" (probably still coping the .gitignore's to the package, which still needs fixing, but the editable installs should no longer be broken).

vyasr:
I hadn't seen the latest release, thanks for the update. Yes, I just tested and it works correctly now.

PR #447, opened 2023-07-31
feat: add a schema for validate-pyproject

henryiii:
Adding a schema, generation from our dataclass-based config system, and the expected plugin.

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -66,7 +66,7 @@ def generate_schema(session: nox.Session) -> None:
     schema = session.run(
         "python", "-m", "scikit_build_core.settings.skbuild_schema", silent=True
     )
-    DIR.joinpath("src/scikit_build_core/resources/toml_schema.json").write_text(schema)  # type: ignore[arg-type]
+    DIR.joinpath("src/scikit_build_core/resources/scikit-build.schema.json").write_text(schema)  # type: ignore[arg-type]
 
 
 @nox.session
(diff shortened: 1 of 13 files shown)


henryiii:
CC @abravalheri if you'd like to take a look. :)

noxfile.py
```diff
@@ -60,6 +60,26 @@ def _run_tests(
...
+
+
+@nox.session
+def validate_schemas(session: nox.Session) -> None:
+    session.install("-e.", "validate-pyproject")
+
+    docs_examples = Path("docs/examples").glob("**/pyproject.toml")
+    tests_packages = Path("tests/packages").glob("**/pyproject.toml")
+    paths = [*docs_examples, *tests_packages]
+    # mop currently required to make mypy happy, future nox version should fix
```
henryiii, line 79:
```suggestion
    # map currently required to make mypy happy, future nox version should fix
```

abravalheri:
Sorry for the delay @henryiii.

I had a look on the parts that I have some knowledge (and I admit I skipped the implementation details to generate a JSON schema 😝), but things look fine to me.

Specifically, I believe that `"validate_pyproject.tool_schema".scikit-build = "scikit_build_core.settings.skbuild_schema:get_skbuild_schema"` should work fine with `validate-pyproject`.

If you want to add any integration test something you can try is ([ref](https://github.com/abravalheri/validate-pyproject/blob/main/tests/test_api.py#L96C1-L103C44)):

```python
    def test_valid(valid_example: dict):
        validator = validate_pyproject.api.Validator()
        assert validator(valid_example) is not None

    def test_invalid(invalid_example: dict):
        validator = validate_pyproject.api.Validator()
        with pytest.raises(Exception):
            validator(invalid_example)
```

abravalheri:
`sphinx-jsonschema` is the best I could find in `sphinx` space for JSON schemas, but sometimes it still can be difficult to read the generated HTML.

There might be better alternatives in the `javascript` universe, but that is difficult to integrate with sphinx or with `pip install`[^1].

[^1]: It there is something I hate doing is setting up a nodejs environment in the CI just to be able to run a couple nodejs-based CLI tools...

PR #449, opened 2023-08-02
ci: fix fedora tests

henryiii:
Fix the fedora tests.


henryiii:
Is there a way to disable the fortran example for Python 3.12? NumPy does not support Python 3.12, that's what's breaking the remaining two jobs.

PR #453, opened 2023-08-08
Fix error in config names

jpn--:
Self explanatory.  I followed the migration guide as written and got an error.


henryiii (approved):

henryiii:
Thanks!

Issue #435, opened 2023-07-17
Extract version from source file?


ChrisBarker-NOAA:
@henryiii told me that being able to extract the version from the source was on the list, but not yet done.

But I don't see an issue on it.

How can I help? It seems like pretty straightforward code to write, once I know where to put it.

NOTE: I'd like to see two options:

1) setuptools-like version from a code file:

`version = {attr = "my_package.VERSION"}`

2) setuptools_scm style: writing to a dynamically generated file:

`write_to = "src/binary4fun/_version.py"`

In this case, unlike setuptools_scm, I'd like to be able to be flexible about where the version comes from -- e.g. it could be extracted from SCM, or simply hard-coded in the pyproject.toml

bennyrowland:
Hi Chris,
This seems like two orthogonal suggestions to me:
1) determining the version
2) outputting the determined version to a file

We already support running `setuptools_scm` itself (see https://github.com/scikit-build/scikit-build-core/blob/main/docs/configuration.md#dynamic-metadata).

The way to handle a new dynamic version approach would be to create a new dynamic metadata plugin (whether this would be part of `scikit-build-core` or a separate package would be up to @henryiii). This could look like (in `pyproject.toml`):

```
[tool.scikit-build]
metadata.version.provider = "metadata_plugin"
metadata.version.attr = "my_package.VERSION"
```

or alternatively

```
[tool.scikit-build]
metadata.version.provider = "metadata_plugin"

[tool.metadata-plugin]
attr = "my_package.VERSION"
```
if you wanted to separate out the config for the plugin.

Generating a file containing the version info in some way is really separate from working out what the version should be. `setuptools_scm` bundles the two components together, and so another plugin could do the same, but maybe there is some value in having a separate component that can write the version into a file independently of how the version is worked out. One simple way for end users to do this without changing `scikit-build-core` would be to do it in CMake, using the `SKBUILD_PROJECT_VERSION` variable which is injected into CMake by `scikit-build-core`. Writing that into a template file or similar would be very easy and flexible. Adding something to the docs on how to do that could be useful.

ChrisBarker-NOAA:
> determining the version
> outputting the determined version to a file

Absolutely -- definitely want to separate those!

> setuptools_scm bundles the two components together

it bundles everything together -- I don't think we should follow that model :-)

> a separate component that can write the version into a file independently of how the version is worked out.

yes!

[...]

PR #457, opened 2023-08-10
feat: regex and dynamic-metadata rewrite

henryiii:
Based on https://github.com/scikit-build/dynamic-metadata/pull/1 (with some updates). Fixes https://github.com/scikit-build/scikit-build-core/issues/435.

This doesn't start using `tool.dynamic-metadata` yet, to give us room to change that. Plus we have to back-compat support `tool.scikit-build.metadata`, so might as well keep that as the only way to set this for now.

### DIFF
--- a/docs/api/scikit_build_core.metadata.rst
+++ b/docs/api/scikit_build_core.metadata.rst
@@ -17,6 +17,14 @@ scikit\_build\_core.metadata.fancy\_pypi\_readme module
    :undoc-members:
    :show-inheritance:
 
+scikit\_build\_core.metadata.regex module
+-----------------------------------------
+
+.. automodule:: scikit_build_core.metadata.regex
+   :members:
+   :undoc-members:
+   :show-inheritance:
+
 scikit\_build\_core.metadata.setuptools\_scm module
 ---------------------------------------------------
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -445,6 +445,31 @@ metadata.readme.provider = "scikit_build_core.metadata.fancy_pypi_readme"
 
 :::
 
+:::{tab} Regex
+
+If you want to pull a string-valued expression (usually version) from an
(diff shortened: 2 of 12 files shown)


henryiii:
A few thoughts: (A) Should we have a write_to / template pair of options for the regex plugin? (B) Or should we put metadata file writing into scikit-build-core generally? (C) Or should we build in something to dynamic-metadata to handle this generally? One benefit of the last one is that you could probably keep from writing the file out into the source tree - except probably with editable installs.

I think I'm pretty strongly in favor of B, maybe along with a mention in dynamic-metadata of how backends could do this.

henryiii:
@bennyrowland - I'm making this small change to the interface for plugins following https://github.com/scikit-build/dynamic-metadata/pull/1. Let me know if there are any issues. I've been moving ahead with option B above, so will likely merge this soon and then open a PR for option B.

bennyrowland:
@henryiii, sorry for the slow response, sadly my time allocation for responding to OSS issues doesn't align well with your extremely rapid development cycle. But rest assured that I am continuing to follow along with developments as much as I can. This is the change so that each plugin gets called per field, rather than allowing one plugin to provide multiple field values in a single call? I can understand that this makes the API rather simpler/cleaner and in practice it is unlikely to be much of an issue (I guess that in extremis a plugin could cache the results of its expensive run and regurgitate those values in subsequent calls).

I also wanted to say that I really like the new "write metadata to file" implementation, it is very elegant. The only thing I wanted to query was about whether you had any thoughts about alternative templating options (like Jinja or Mako) which could then have more general purpose Python code embedded in them. That would make the `generate` settings into a more versatile file generation option which could also be used for generating things like build logs with timestamps or the values of environment variables, and probably a whole bunch of other use cases. I don't have a particular need for this kind of thing, but it just occurred to me looking at the changes and wondered what your perspective was.

PR #462, opened 2023-08-11
fix: report cmake/ninja requried if already present

### LINKED ISSUE #458: get_requires_for_build_wheel confuses tox4 by not requesting cmake if it is in the build venv
I just moved a project from old `scikit-build` and `tox3` to `scikit-build-core` and `tox4` - the improved caching on both sides has taken my test time from several minutes to several seconds, so very happy about that. Unfortunately, `scikit-build-core`'s clever use of `get_requires_for_build_wheel` to add `cmake` etc. only when they are not already available is backfiring somewhat.

The first run works perfectly, with `tox` installing `cmake` in the build environment. For the second run `scikit-build-core` finds that `cmake` is available in its path and so does not request it in `get_requires_for_build_wheel()`. `tox` sees that the requirements have changed, and scraps the build env (and promptly crashes, but that is a secondary issue, I think). So I think that a better solution would be for `scikit-build-core` to check if `cmake` is available through the env (rather than anywhere else) and if so then it should continue to say that it requires the package. Since the package is already installed, it won't make a material difference to the run time of anything for `pip` to decide it doesn't need to do anything, but it will stop `tox` from getting confused about the changing requirements. Note that if I add `cmake` to the list of build requirements in `pyproject.toml` (so `scikit-build-core` never asks for it and never changes its mind) then the problem goes away.

I guess that there are some technical questions about deciding whether `cmake` is coming from the environment, but the simplest way is probably just to `import cmake` and check for an error. This could be a short-circuit at the top of GetRequires.cmake(). If `cmake` can be imported then `yield f"cmake>={cmake_min}"`. @henryiii I am happy to submit a PR if this change meets with your approval.

henryiii:
Fix #458.

@bennyrowland, I think this either fixes this or is a good start.

### DIFF
--- a/src/scikit_build_core/builder/get_requires.py
+++ b/src/scikit_build_core/builder/get_requires.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import dataclasses
 import functools
+import importlib.util
 import os
 import sysconfig
 from collections.abc import Generator, Mapping
@@ -55,6 +56,13 @@ class GetRequires:
 
     def cmake(self) -> Generator[str, None, None]:
         cmake_min = self.settings.cmake.minimum_version
+
+        # If the module is already installed (via caching the build
+        # environment, for example), we will use that
+        if importlib.util.find_spec("cmake") is not None:
+            yield f"cmake>={cmake_min}"
+            return
+
         cmake = best_program(
             get_cmake_programs(module=False), minimum_version=cmake_min
         )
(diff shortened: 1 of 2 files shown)


bennyrowland (approved):
Running this version fixes running my project via `tox`, so I am very happy with it. I note that there aren't any additional tests to check for this behaviour - is this something that is worth adding, or are you happy that it is a small enough change. Adding a test would be quite challenging with the current changes to `protect_get_requires()` as that explicitly always says that `cmake`/`ninja` are not installed (and is `autouse`d), so we can't test that they are still requested when they are installed.

henryiii:
I'm okay with it for now, I think (especially since I saw it broke the tests as expected without the modification above).

PR #485, opened 2023-09-05
fix: fix template-file with template-path

Freed-Wu:
See https://github.com/scikit-build/scikit-build-core/blob/f5bad6241069bd09b59e64befb93e78a0d0c907d/src/scikit_build_core/settings/skbuild_read_settings.py#L123


henryiii:
Thanks!

Issue #500, opened 2023-09-17
 scikit_build_core.errors.CMakeNotFoundError: Could not find CMake with version >= 3.12


DaveScream:
`C:\Users\F2>cmake --version
cmake version 3.26.4-msvc4

CMake suite maintained and supported by Kitware (kitware.com/cmake).

C:\Users\F2>where cmake
C:\Program Files\Microsoft Visual Studio\2022\Community\Common7\IDE\CommonExtensions\Microsoft\CMake\CMake\bin\cmake.exe

C:\Users\F2>python -m pip install llama-cpp-python --force-reinstall --upgrade --no-build-isolation -v
Using pip 23.0.1 from C:\Python310\lib\site-packages\pip (python 3.10)
Looking in indexes: https://pypi.org/simple, https://pypi.ngc.nvidia.com
Collecting llama-cpp-python
  Using cached llama_cpp_python-0.2.6.tar.gz (1.6 MB)
  Running command Preparing metadata (pyproject.toml)
  2023-09-17 22:51:22,033 - scikit_build_core - WARNING - Could not determine CMake version, got 'cmake version 3.26.4-msvc4\n\nCMake suite maintained and supported by Kitware (kitware.com/cmake).\n'
  Traceback (most recent call last):
    File "C:\Python310\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 353, in <module>
      main()
    File "C:\Python310\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 335, in main
      json_out['return_val'] = hook(**hook_input['kwargs'])
    File "C:\Python310\lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 149, in prepare_metadata_for_build_wheel
      return hook(metadata_directory, config_settings)
    File "C:\Python310\lib\site-packages\scikit_build_core\build\__init__.py", line 73, in prepare_metadata_for_build_wheel
      return _build_wheel_impl(
    File "C:\Python310\lib\site-packages\scikit_build_core\build\wheel.py", line 106, in _build_wheel_impl
      cmake = CMake.default_search(minimum_version=settings.cmake.minimum_version)
    File "C:\Python310\lib\site-packages\scikit_build_core\cmake.py", line 49, in default_search
      raise CMakeNotFoundError(msg)
  scikit_build_core.errors.CMakeNotFoundError: Could not find CMake with version >= 3.12
[...]

PR #508, opened 2023-09-20
Parse CMake version strings containing '-'

jllllll:
Uses a simple `str.split('-msvc')` to extract the version number. `'-msvc'` in particular was chosen to allow for forwards-compatibility with future builds of MSVC CMake. I did not check for older versions as I do not expect they will be any less compatible and this is still an improvement on the current situation of no compatibility at all.

In my opinion, splitting on `-` should be sufficient and provide compatibility with any other builds of CMake with custom version strings. However, I did not do this as I'm not familiar with all of the possible builds of CMake out there and wanted to ensure functionality.

---
Update: Now parses using `str.split('-')` for extended compatibility. Thanks @henryiii 

---
Resolves: #500

### DIFF
--- a/src/scikit_build_core/program_search.py
+++ b/src/scikit_build_core/program_search.py
@@ -78,7 +78,7 @@ def get_cmake_programs(*, module: bool = True) -> Generator[Program, None, None]
 
         try:
             version = Version(
-                result.stdout.splitlines()[0].split()[-1].split("-msvc")[0]
+                result.stdout.splitlines()[0].split()[-1].split("-")[0]
             )
         except (IndexError, InvalidVersion):
             logger.warning(f"Could not determine CMake version, got {result.stdout!r}")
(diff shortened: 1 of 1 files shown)


src/scikit_build_core/program_search.py
```diff
@@ -77,7 +77,9 @@ def get_cmake_programs(*, module: bool = True) -> Generator[Program, None, None]
             continue
 
         try:
-            version = Version(result.stdout.splitlines()[0].split()[-1])
+            version = Version(
+                result.stdout.splitlines()[0].split()[-1].split("-msvc")[0]
```
henryiii, line 81:
```suggestion
                result.stdout.splitlines()[0].split()[-1].split("-")[0]
```

I think it's fine to always remove any extension here. These are usually used to indicate extra features or a build number, which shouldn't cause it to be invalid.

henryiii (approved):

PR #526, opened 2023-10-17
Overrides platform node

0xTowel:
This adds support for `platform_node` in the overrides. Also takes care of a small typo in the markdown syntax for the main project README.


henryiii:
Thanks!

henryiii:
(Let me know if there are any others that would make sense to have. I wouldn't have thought `platform.node` was useful)

PR #530, opened 2023-10-20
Fix bad config option name in configuration.md

0xTowel:
The system platform is referenced as `platform_system` in the toml, not `sys_platform`. It is still computed by `sys.platform` however.


henryiii:
Thanks! Actually, aren't these all dashes instead of underscores?

PR #564, opened 2023-12-04
feat: Preserve additivity of `cmake.define` across `overrides` tables

stubbiali:
This light-weight PR adds a special rule when overriding `cmake.define`, so that additivity is retained and orthogonal CMake defines are accommodated. Consider the following TOML excerpt:

```toml
if.env.SET_FOO = "ON"
cmake.define.FOO = "1"

if.env.SET_BAR = "ON"
cmake.define.BAR = "1"
```

As of now, this would result in `cmake.define` being `{"BAR": "1"}`. With the PR, `cmake.define` will be `{"FOO": 1, "BAR": "1"}`.

### DIFF
--- a/src/scikit_build_core/resources/scikit-build.schema.json
+++ b/src/scikit_build_core/resources/scikit-build.schema.json
@@ -436,6 +436,63 @@
               }
             ]
           },
+          "inherit": {
+            "type": "object",
+            "properties": {
+              "cmake": {
+                "type": "object",
+                "additionalProperties": false,
+                "properties": {
+                  "args": {
+                    "$ref": "#/$defs/inherit"
+                  },
+                  "define": {
+                    "$ref": "#/$defs/inherit"
+                  },
+                  "targets": {
+                    "$ref": "#/$defs/inherit"
+                  }
+                }
+              },
+              "sdist": {
+                "type": "object",
+                "additionalProperties": false,
+                "properties": {
+                  "include": {
+                    "$ref": "#/$defs/inherit"
+                  },
+                  "exclude": {
+                    "$ref": "#/$defs/inherit"
+                  }
+                }
+              },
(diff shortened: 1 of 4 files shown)


henryiii:
I've been thinking about this (also in cibuildwheel) for quite a while, but since there's no "null" option in TOML, there would be no way to un-set an environment variable once set. It's also inconsistent with how other options work. What I've thought of is having a special character/name to indicate that it's going to be additive. Probably "cmake.define+`?

henryiii:
(Haven't thought through this at all yet, but how about something like `{inherit = true}` as an item in the list filling in the old version when it's encountered?)

stubbiali:
What about `'+cmake.define.FOO' = "1"`? Of course, parsing of the keys will be needed. But this approach could be applied to `cmake.args` too, to append items to the list.

stubbiali:
The logic to support the syntax `'+cmake.args'` and `'+cmake.define'` has been implemented.

henryiii:
(This is moving forward slowly, I'll get back to you once we are happy on a design! I've brought it up with the rest of the cibuildwheel devs at https://github.com/pypa/cibuildwheel/discussions/1716)

henryiii:
If people like https://github.com/pypa/cibuildwheel/pull/1730, we can do the same thing here.

stubbiali:
Speaking on behalf of the team behind https://github.com/ghex-org/GHEX: we like the approach pursued in https://github.com/pypa/cibuildwheel/pull/1730, and the end result will perfectly serve our needs, so green light from our side. If needed, I can gladly offer some help.

henryiii:
There's some question about the final term ("inherit" vs. "extend", opinions welcome on that issue), but it's mostly finalized. I can look at adding it here too.

PR #565, opened 2023-12-04
feat: Setting CMake variables though environment variables

stubbiali:
This PR allows reading environment variables to define CMake variable. This is especially useful when the set of valid values which a CMake variable can assume is either not known a priori, or is not finite, and so `overrides` is not really helpful. 

Simple example:

```toml
[tool.scikit-build]
cmake.define.FOO = {env = "FOO"}
```

Bumps `typing_extensions` to being required for Python 3.9+.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -386,6 +386,14 @@ SKBUILD_CMAKE_DEFINE: SOME_DEFINE=ON
 
 ````
 
+You can also (`pyproject.toml` only) specify a dict, with `env=` to load a
+define from an environment variable, with optional `default=`.
+
+```toml
+[tool.scikit-build.cmake.define]
+SOME_DEFINE = {env="SOME_DEFINE", default="EMPTY"}
+```
+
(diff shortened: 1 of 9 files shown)


henryiii:
What about `cmake.define.FOO = {env = "FOO"}`?

stubbiali:
I'm pretty open about the syntax to use. My suggestion were inspired by f-strings in Python, but I also like the idea of having a dictionary. Do you think that specifying a default value for when the env variable is not set might be useful? Something like `cmake.define.FOO = {env = "FOO", default = 0}`.

stubbiali:
The logic to support the syntax `cmake.define.FOO = {env = "FOO", default = 0}` has been implemented.

henryiii:
FYI, I'm taking a look at this slowly, playing with some ideas to integrate it a bit more fully (we'll want it to be part of the schema too). Should hopefully get somewhere later this week.

henryiii (approved):

henryiii:
Thanks!

PR #568, opened 2023-12-07
docs: fix a link

henryiii:
Fix a link & tiny cleanup.


henryiii:
Ahh, it looks likes links were broken by https://github.com/pypa/packaging.python.org/pull/1396. I'll update them for now, later we could probably use intersphinx (though the labels seem to be changed too).

PR #626, opened 2024-02-02
fix: do not exit with exec permission error in ninja check

thewtex:
Also `cmake` check.

As discussed in:

  https://github.com/orgs/scikit-build/discussions/1057

There may be times when we attempt to execute `ninja --version` from the ninja package in /tmp/, which may be mounted `noexec`. This result in a PermissionError.

Carry on so we can inspect other ninja executables available.


henryiii (approved):

henryiii:
Released 0.8.1.

PR #638, opened 2024-02-14
Fix '-' vs '_' in error message

nbelakovski:
The necessary setting is `build-dir` not `build_dir`

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -566,7 +566,7 @@ Known limitations:
 
 ```console
 # Very experimental rebuild on initial import feature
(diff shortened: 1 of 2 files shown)


docs/configuration.md
```diff
@@ -566,7 +566,7 @@ Known limitations:
 
 ```console
 # Very experimental rebuild on initial import feature
-$ pip install --no-build-isolation --config-settings=editable.rebuild=true -ve.
+$ pip install --no-build-isolation --config-settings=editable.rebuild=true --config-settings=build-dir=/path/to/my/build/dir -ve.
```
henryiii, line 569:
```suggestion
$ pip install --no-build-isolation --config-settings=editable.rebuild=true -Cbuild-dir=build -ve.
```

henryiii:
Thanks!

Issue #642, opened 2024-02-19
Specifying `editable.mode=inplace` when `build-dir` already set in `pyproject.toml` is not supported


jcfr:
This issue is related to the `inplace` experimental mode associated with editable install.

Links:
* https://scikit-build-core.readthedocs.io/en/latest/configuration.html#editable-installs
* https://github.com/scikit-build/scikit-build-core/pull/553

henryiii:
Ahh, okay. You can set the build-dir to “” until this is fixed.

PR #643, opened 2024-02-19
doc: Clarify requirement for using "inplace" editable mode

jcfr:
Until #642 is addressed, specifying `-Cbuild-dir=""` is required to avoid "build-dir must be empty for editable inplace mode" error.


henryiii (approved):
I think we can fix this pretty quickly, but docs updates are even quicker. :)

PR #648, opened 2024-02-23
Fix __path__ with editable installs containing symlinks

vyasr:
TODO (currently just a test)


vyasr:
Going to close this PR for now since it was just to demonstrate an issue. Happy to reopen if we come up with a solution and if I end up working on it myself.

PR #652, opened 2024-02-26
fix!: exclude installed files if listed in exclude

henryiii:
As mentioned in our recent Community Meeting.

### DIFF
--- a/tests/test_simplest_c.py
+++ b/tests/test_simplest_c.py
@@ -82,6 +82,7 @@ def test_pep517_wheel(tmp_path, monkeypatch, virtualenv, component):
 
         if not component:
             expected_wheel_files.add("generated_ignored.txt")
+            expected_wheel_files.add("generated_no_wheel.txt")
 
         if not component or "Generated" in component:
             expected_wheel_files.add("generated.txt")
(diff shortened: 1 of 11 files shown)


tests/test_simplest_c.py
```diff
@@ -110,7 +111,10 @@ def test_pep517_wheel_incexl(tmp_path, monkeypatch, virtualenv):
         {
             "sdist.include": "src/simplest/*included*.txt",
             "sdist.exclude": "src/simplest/*excluded*.txt",
-            "wheel.exclude": "src/simplest/sdist_only.txt",
+            "wheel.exclude": [
+                "simplest/sdist_only.txt",
+                "simplest/generated_no_wheel.txt",
+            ],
```
henryiii, line 117: This is a behavior change, though I'd be willing to consider the old behavior a bug. Previously, wheel exclude matched on the SDist/source path. Now it matches on the final wheel path.

Not sure to just call this a bug fix, or to add back-compat for it via `minimum-version`.

henryiii: No packages (by either PyPI search or GitHub search) were found that would be affected by this change, so the plan is to make it as a bug fix, document this in the changelog, and be open to adding a compat setting for it if someone does end up needing it.

PR #655, opened 2024-02-28
fix: Ensure wheel.platlib being false sets <abi>-<platform> as "none-any"

jcfr:
* [x] Add regression test or update existing one


jcfr:
This will be rebased once #657 is integrated :rocket:

jcfr:
@henryiii This is pending review & approval :pray:

Once this is integrated, what do you think of releasing `0.8.2` ?

henryiii (approved):

henryiii:
Very soon, a few more small fixes, but probably can plan on one by the weekend.

jcfr:
In the meantime, thanks for the review

henryiii:
Pushed 0.8.2.

PR #676, opened 2024-03-28
docs: improve ninja.make-fallback description in the README

thewtex:


thewtex:
@henryiii I presume this is the right description.

henryiii:
This comes from `src/scikit_build_core/settings/skbuild_model.py`. The other locations can be regenerated with `nox -s readme generate_schema`.

henryiii (approved):

PR #698, opened 2024-04-11
fix: ensure many/musl tags not selected

henryiii:
This was tagging as manylinux (at least sometimes).

### DIFF
--- a/src/scikit_build_core/builder/wheel_tag.py
+++ b/src/scikit_build_core/builder/wheel_tag.py
@@ -48,7 +48,14 @@ class WheelTag:
                 msg = f"Unexpected build-tag, {build_tag!r} cannot contain dashes"
                 raise AssertionError(msg)
 
-        best_tag = next(packaging.tags.sys_tags())
+        # manylinux sometimes comes before linux, but can't assume manylinux, use auditwheel instead
+        best_tag = next(
+            iter(
+                p
+                for p in packaging.tags.sys_tags()
+                if "manylinux" not in p.platform and "muslllinux" not in p.platform
+            )
+        )
         interp, abi, *plats = (best_tag.interpreter, best_tag.abi, best_tag.platform)
         pyvers = [interp]
 


mryzhov:
Hi @henryiii, looks like after this change it would not be possible to set "manylinux" platform tag during the wheel packaging.
Imho, the previous variant was more flexible (

henryiii:
It was not correct. `manylinux` means you have passed `auditwheel`. Tools like `cibuildwheel` run `auditwheel` for you. But without that, the wheel is _not_ `manylinux`, that is, it's not redistributable, it is only guaranteed to work on your machine and nowhere else.

The fix is to run `auditwheel` on your wheel before uploading them.

henryiii:
This is the bug in packaging: https://github.com/pypa/packaging/issues/160

It was sadly closed in favor of a new tag "local" that never went anywhere. Producing a manylinux/musllinux by default is clearly incorrect.

mryzhov:
> This is the bug in packaging: [pypa/packaging#160](https://github.com/pypa/packaging/issues/160)
> 
> It was sadly closed in favor of a new tag "local" that never went anywhere. Producing a manylinux/musllinux by default is clearly incorrect.

Hi @henryiii , thank you for the answer. You are right that default value should not be a manylinux. But by this change you disabled an ability to set this tag manually at all. We build wheels which satisfy manylinux requirements and using additional tool to change the tag only is an unnecessary dependency for us( 
It would be great to have a configuration or environment variable to set it during the packaging manually.

henryiii:
No other build backends (such as setuptools) has a way to configure this. It's very dangerous; a broken "manylinux" wheel will always work on the system it was built on; it also may work on some other systems. It's very hard to tell without validating it, which is what auditwheel does. It also depends on where you build the wheel; if you are using the manylinux docker container, then maybe you are producing manylinux-compliant wheels already, but if someone builds on a normal OS, it might not be compliant. The only way to be sure is to verify with auditwheel.

Also, you are producing a manylinux_2_27 tag, what is that? Manylinux docker images come in 2_24 (abandoned) and 2_28, not 2_27. And manylinux2014 is the standard tag name for 2_17, but this is grabbing the less-compatible "modern" name accidentally. Also, when I run auditwheel on your published wheel, I get "libopenvino" can't be found; checking, it's actually inside a different wheel, openvino! This is not allowed by manylinux. All wheels must bundle their own dependencies and name mangle them so they can be safely loaded with other wheels that have different requirements on the lib. Now, if you specifically require exactly one version of `openvino`, and are pretty sure no one else will use libopenvino, you can make it work, but I don't think this should be "automatic", as you are hiding a potentially problematic design. Instead, you can do:

```console
$ # I'm starting from PyPI, you don't need these steps
$ docker run --rm -it quay.io/pypa/manylinux2014_x86_64
# python3.12 -m pip download openvino-tokenizers
Saved /openvino_tokenizers-2024.1.0.0-81-py3-none-manylinux_2_17_x86_64.whl
Saved /openvino-2024.1.0-15008-cp312-cp312-manylinux2014_x86_64.whl
Saved /numpy-1.26.4-cp312-cp312-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
Saved /openvino_telemetry-2024.1.0-py3-none-any.whl
Saved /packaging-24.0-py3-none-any.whl
Successfully downloaded openvino-tokenizers openvino numpy openvino-telemetry packaging
# pipx install wheel
  installed package wheel 0.43.0, installed using Python 3.10.14
  These apps are now globally available
    - wheel
done! ✨ 🌟 ✨
# wheel tags --remove --platform-tag linux_x86_64 openvino_tokenizers-2024.1.0.0-81-py3-none-manylinux_2_17_x86_64.whl
openvino_tokenizers-2024.1.0.0-81-py3-none-linux_x86_64.whl

# # Now we have the current output, a linux wheel
# # This is the line you need
# auditwheel repair openvino_tokenizers-2024.1.0.0-81-py3-none-linux_x86_64.whl --exclude libopenvino_tensorflow_frontend.so.2410 --exclude libopenvino.so.2410
INFO:auditwheel.main_repair:Repairing openvino_tokenizers-2024.1.0.0-81-py3-none-linux_x86_64.whl
INFO:auditwheel.lddtree:Excluding libopenvino_tensorflow_frontend.so.2410
INFO:auditwheel.lddtree:Excluding libopenvino.so.2410
INFO:auditwheel.wheeltools:Previous filename tags: linux_x86_64
INFO:auditwheel.wheeltools:New filename tags: manylinux_2_17_x86_64, manylinux2014_x86_64
INFO:auditwheel.wheeltools:Previous WHEEL info tags: py3-none-linux_x86_64
INFO:auditwheel.wheeltools:New WHEEL info tags: py3-none-manylinux_2_17_x86_64, py3-none-manylinux2014_x86_64
INFO:auditwheel.main_repair:
Fixed-up wheel written to /wheelhouse/openvino_tokenizers-2024.1.0.0-81-py3-none-manylinux_2_17_x86_64.manylinux2014_x86_64.whl
```

As an added bonus, notice the wheel is also tagged with the compat tag `manylinux2014_x86_64` too!

mryzhov:
> ```
> python3.12 -m pip download openvino-tokenizers
> ```
Auditwheel is really useful to verify the manylinux wheel, but I would disagree that this is only one way to build correct manylinux wheel. The provided solution require to manually exclude some libraries(`--exclude libopenvino_tensorflow_frontend.so.2410 --exclude libopenvino.so.2410`) and we would need to change it every release, in addition to that auditwheel changes the initial wheel structure (added empty openvino_tokenizers.libs dir).

PR #725, opened 2024-04-24
ci: macos-latest is changing to macos-14 ARM runners

henryiii:
Committed via https://github.com/asottile/all-repos

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -281,7 +281,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -281,7 +281,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        runs-on: [ubuntu-latest, macos-latest, windows-latest]
+        runs-on: [ubuntu-latest, macos-13, windows-latest]
```
henryiii, line 284:
```suggestion
        runs-on: [ubuntu-latest, macOS-14, windows-latest]
```

PR #739, opened 2024-05-09
docs: fix example of configuration overrides in configuration.md

wu-vincent:
The example provided is not working anymore and will yield errors. This pull request will fix the issue.


henryiii (approved):
Thanks!

PR #744, opened 2024-05-14
chore: pin pylint for now

henryiii:
Pylint 3.2.0 reports an error in a .pyi file that 3.1.* didn't; we are ignoring pyi files since 3.0 since pylint doesn't handle them, but pylint is now ignoring the ignore. https://github.com/pylint-dev/pylint/issues/9623 (See boost-histogram if you want see how many errors pylint can spew about pyi files! We just have one.)


henryiii:
The format job passes, so going to merge to unblock the CI on other PRs.

PR #748, opened 2024-05-16
Strip whitespace around generator so that it's passed through correctly

vyasr:
It looks like without this change, the leading whitespace with something like `-G 'Unix Makefiles'` being converted to ` 'Unix Makefiles'` makes CMake misinterpret this and not realize a generator is being provided.


henryiii (approved):

henryiii:
Thanks!

PR #754, opened 2024-05-23
fix(docs): note that -C options can be prefixed with `skbuild.`

KyleFromNVIDIA:
https://github.com/scikit-build/scikit-build-core/pull/556 added the ability for -C options to be prefixed with `skbuild.`, but this was not noted in the configuration documentation. Add a note for this.


henryiii (approved):

henryiii:
Thanks! This is mostly for allowing user options in the future, which we haven't added yet.

KyleFromNVIDIA:
I noticed this when working on https://github.com/rapidsai/rapids-build-backend/pull/37. In general I think namespacing config options is a good idea - we plan to do it for rapids-build-backend.

PR #762, opened 2024-06-05
fix: improve error message when metadata.version.provider not set

thewtex:
When using setuptools_scm to obtain the version dynamic and

```toml
[tool.scikit-build]
metadata.version.provider = "scikit_build_core.metadata.setuptools_scm"
```

is not set in pyproject.toml, the error message suggested that only a
static version is supported. Modify the error message to indicate both
are supported and suggest a fix in the dynamic case.


henryiii:
It was probably true when written. :)

henryiii (approved):

PR #768, opened 2024-06-14
redesign cmake entrypoint

kiwifb:
This is a sort of follow up to #413 and #414. In sage-on-gentoo we discovered that the presence of scikit-build-core breaks the building sagemath (https://github.com/cschwan/sage-on-gentoo/issues/783). scikit-build-core also happens to be an indirect dependency for building sagemath's documentation.
On the basis that the presence of setuptolls_rust does not break sagemath's build, I studied the difference with scikit-build-core. I noticed that the elements `bundled in the single "finalize_distribution_options" entrypoint in scikit is split between two different setuptools entrypoints.
This PR reproduce that setup and does a bit of clean up. The bundling of the old _prepare_extension_detection and _prepare_build_cmake_command into finalize_distribution_options being redundant after attributing them directly to separate entrypoints. I also renamed the functions to better characterise them.

Using this PR, sagemath builds without problems and pyzmq, which uses scikit-build-core also builds properly.


src/scikit_build_core/setuptools/build_cmake.py
```diff
@@ -192,7 +192,7 @@ def __len__(self) -> int:
         dist.ext_modules = getattr(dist, "ext_modules", []) or EvilList()
 
 
-def _prepare_build_cmake_command(dist: Distribution) -> None:
+def cmake_command(dist: Distribution) -> None:
```
henryiii, line 195:
```suggestion
def finalize_distribution_options(dist: Distribution) -> None:
```

pyproject.toml
```diff
@@ -114,7 +114,8 @@ Issues = "https://github.com/scikit-build/scikit-build-core/issues"
 "distutils.commands".build_cmake = "scikit_build_core.setuptools.build_cmake:BuildCMake"
 "distutils.setup_keywords".cmake_source_dir = "scikit_build_core.setuptools.build_cmake:cmake_source_dir"
 "distutils.setup_keywords".cmake_args = "scikit_build_core.setuptools.build_cmake:cmake_args"
-"setuptools.finalize_distribution_options".scikit_build_entry = "scikit_build_core.setuptools.build_cmake:finalize_distribution_options"
+"distutils.setup_keywords".cmake_extension = "scikit_build_core.setuptools.build_cmake:cmake_extension"
+"setuptools.finalize_distribution_options".scikit_build_entry = "scikit_build_core.setuptools.build_cmake:cmake_command"
```
henryiii, line 118:
```suggestion
"setuptools.finalize_distribution_options".scikit_build_entry = "scikit_build_core.setuptools.build_cmake:finalize_distribution_options"
```

henryiii (commented):
I think it might be better to match the entry points and names?

src/scikit_build_core/setuptools/build_cmake.py
```diff
@@ -19,7 +19,7 @@
 
     from .._compat.typing import Literal
 
-__all__ = ["BuildCMake", "cmake_source_dir", "finalize_distribution_options"]
+__all__ = ["BuildCMake", "cmake_command", "cmake_extension", "cmake_source_dir"]
```
henryiii, line 22:
```suggestion
__all__ = ["BuildCMake", "cmake_extension", "cmake_source_dir", "finalize_distribution_options"]
```

pyproject.toml
kiwifb: This is not compulsory as far as I can see. But I'll follow any convention you want.

src/scikit_build_core/setuptools/build_cmake.py
kiwifb: Of course it would match here. I totally did not pay attention to that line in my initial patch.

kiwifb: Are you OK with me using github "commit suggestion" button or do you prefer a manual process?

src/scikit_build_core/setuptools/build_cmake.py
```diff
@@ -176,7 +176,7 @@ def _has_cmake(dist: Distribution) -> bool:
     )
 
 
-def _prepare_extension_detection(dist: Distribution) -> None:
+def cmake_extension(dist: Distribution) -> None:
```
henryiii, line 179: I am a little confused here, though. Someone doesn't have to specify `cmake_extensions=`, it's not a keyword argument in our API.

However, we do require at least one `cmake_` keyword to be passed. Those could be what ensures this gets called.

(Also, setuptools-rust is a great example and maybe we should check it again to see if matching it's design more closely is a good idea)

We do need to be backward compatible with scikit-build (classic), but we have a wrapper there and we can add whatever we put here there, so that's not really an issue.

Very soon I'm pulling this out into a separate package so it will stop being an issue for modern builds. :)

henryiii: Does setuptools-rust not require a setup.py now?

kiwifb: > Does setuptools-rust not require a setup.py now?

I must say I have no idea on that particular.

src/scikit_build_core/setuptools/build_cmake.py
henryiii: Sure, I’ll squash at the end.

henryiii: Though I think we’re going to have to try slightly different approach, basically by adding it to every keyword argument, rather than adding a new keyword argument that we don’t support (yet?).

src/scikit_build_core/setuptools/build_cmake.py
kiwifb: > I am a little confused here, though. Someone doesn't have to specify `cmake_extensions=`, it's not a keyword argument in our API.
> 
> However, we do require at least one `cmake_` keyword to be passed. Those could be what ensures this gets called.
> 

I only know that passing everything into `finalize_distribution_options` leads to breakage because, in the case of sagemath it ends being used in a place it was never meant to be. And I followed setuptools-rust example as to what the appropriate entrypoint for it should be.

> (Also, setuptools-rust is a great example and maybe we should check it again to see if matching it's design more closely is a good idea)
> 
> We do need to be backward compatible with scikit-build (classic), but we have a wrapper there and we can add whatever we put here there, so that's not really an issue.
> 
> Very soon I'm pulling this out into a separate package so it will stop being an issue for modern builds. :)

I suspected that would happen sooner rather than later. This looks like something that should be named setuptools-cmake.

pyproject.toml
henryiii: IMO it reduces having to look this up, as the only way this is called is via entry point.

src/scikit_build_core/setuptools/build_cmake.py
henryiii: > Does setuptools-rust not require a setup.py now?

Nice:

https://github.com/PyO3/setuptools-rust/blob/34c5f160a3a282ea568d0d285957edbdd927a9c7/pyproject.toml#L39

https://github.com/PyO3/setuptools-rust/blob/34c5f160a3a282ea568d0d285957edbdd927a9c7/setuptools_rust/setuptools_ext.py#L303-L314

kiwifb: > > Does setuptools-rust not require a setup.py now?
> 
> Nice:
> 
> https://github.com/PyO3/setuptools-rust/blob/34c5f160a3a282ea568d0d285957edbdd927a9c7/pyproject.toml#L39
> 
> https://github.com/PyO3/setuptools-rust/blob/34c5f160a3a282ea568d0d285957edbdd927a9c7/setuptools_rust/setuptools_ext.py#L303-L314

I get what you mean more clearly now. Indeed.

henryiii:
Could you try it with my changes and see if that fixes sagemath?

henryiii:
(pyzmq doesn't use the setuptools plugin, so it should be fine. I think we only have ~1 user so far of the setuptools plugin.)

kiwifb:
I just checked and sagemath still builds with your newest changes. I must say, I was not sure it would.

henryiii:
This makes sure nothing is injected unless a `cmake_*` keyword is present. Sagemath doesn't have one of those, so it's fine. :)

The injection should have been pretty safe, but this is safer.

henryiii:
Thanks!

PR #809, opened 2024-07-14
docs: use modern cmake/ninja version setting in Fortran example

henryiii:
Fixing the example.

### DIFF
--- a/docs/examples/getting_started/fortran/pyproject.toml
+++ b/docs/examples/getting_started/fortran/pyproject.toml
@@ -8,5 +8,5 @@ version = "0.0.1"
 dependencies = ["numpy"]
 
 [tool.scikit-build]
-ninja.minimum-version = "1.10"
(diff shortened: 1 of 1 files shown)


flex0902:
The latest pysoild 0.3.3 using this example for compiling fortran program "solid.for".
when I use "pip install pysolid", the error message in cmake is below:
image
The problem seeems to be "\" in the path.
How do I solve this issue.
Many thanks

henryiii:
We haven’t run this example on Windows (requires a Fortran compiler), so it looks like it’s missing a conversion to CMake path.

```cmake
file(TO_CMAKE_PATH "${F2PY_INCLUDE_DIR}" F2PY_INCLUDE_DIR)

# or for CMake 3.20+ only
cmake_path(CONVERT ${F2PY_INCLUDE_DIR} TO_CMAKE_PATH_LIST F2PY_INCLUDE_DIR)
```

PR #816, opened 2024-07-19
improve version regex to capture typed version attributes

maxbachmann:
This allows capturing the version when typing the version attribute using `__version__: str = "1.0.1"` as well.


henryiii (approved):
Great, thanks!

henryiii (approved):

PR #854, opened 2024-08-06
Fix crash when building inside a submodule

ausbin:
If .git is a file instead of a directory (such as when you are building in a submodule), scikit-build-core will currently crash because a NotADirectoryError is thrown instead of a FileNotFoundError. This PR accounts for the resulting exception (NotADirectoryError) instead of crashing.

Another option instead of hardcoding this list would be to catch the more general `OSError`. However, I did not do that, because I am worried about silently swallowing too broad of a class of exceptions.

I wrote a unit test and also tested the fix [against the project][1] where I encountered this.

[1]: https://github.com/scikit-build/scikit-build-core/blob/main/.github/CONTRIBUTING.md#testing-a-project-with-a-branch--main

(By the way, this project is very helpful, thank you. The fact that I encountered this issue only a couple of hours after the release of 0.10.0 tells you how often I use it!)

### DIFF
--- a/src/scikit_build_core/build/_file_processor.py
+++ b/src/scikit_build_core/build/_file_processor.py
@@ -13,7 +13,7 @@ if TYPE_CHECKING:
 __all__ = ["each_unignored_file"]
 
 EXCLUDE_LINES = [
-    ".git/",
+    ".git",
     ".tox/",
     ".nox/",
     ".egg-info/",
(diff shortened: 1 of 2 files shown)


src/scikit_build_core/build/_file_processor.py
```diff
@@ -13,7 +13,7 @@
 __all__ = ["each_unignored_file"]
 
 EXCLUDE_LINES = [
-    ".git/",
+    ".git",
```
ausbin, line 16: This is so the `.git` file itself does not show up in the list

henryiii:
I think the error here actually depends on the operating system. https://github.com/python/cpython/issues/119993 PermissionsError is another one.

henryiii (approved):

henryiii:
Thanks for a great and fast PR that included a test! I’ll release once I get #853 working on PyPy.

PR #857, opened 2024-08-12
fix: Files included twice when they match `include_spec`

abetlen:
Currently if a file matches the explicit `sdist.include` it is included twice in the source distribution.

This causes an error with some package managers such as `uv` (See https://github.com/abetlen/llama-cpp-python/issues/1670#issuecomment-2284603172).


henryiii (approved):

henryiii:
Deploying 0.10.2 now, thanks!

PR #872, opened 2024-08-22
fix: sdist.exclude takes precedence over .gitignore

### LINKED ISSUE #871: sdist.exclude should not be overridden by .gitignore
My `.gitignore` file uses the following trick to ignore (most) files without extensions:

```
*
!*.*
!*/
```

Unfortunately, the way that `sdist.exclude` is implemented is by _prepending_ the listed rules to these. These are then overridden by the `*` in the file.

I believe that `sdist.exclude` should be implemented similarly to `sdist.include`; in a separate `pathspec` that is checked after `sdist.include` and which gets final say.

alexreinking:
Previously, sdist.exclude and the built-in EXCLUDE_LINES were prepended to the list of .gitignore rules. However, a common pattern of using * to exclude everything and then using !-rules to list what should be included renders this feature useless.

This commit changes the behavior of each_unignored_file to sequentially consider the following sets of independent rules:

1. Explicit inclusions
2. Explicit exclusions
3. Global exclusions from the .gitignore files
4. The built-in exclusions
5. Nested exclusions from .gitignore

Fixes #871

### DIFF
--- a/src/scikit_build_core/build/_file_processor.py
+++ b/src/scikit_build_core/build/_file_processor.py
@@ -36,12 +36,11 @@ def each_unignored_file(
     """
     Runs through all non-ignored files. Must be run from the root directory.
     """
-    exclude_lines = EXCLUDE_LINES + list(exclude)
-
+    global_exclude_lines = []
     for gi in [Path(".git/info/exclude"), Path(".gitignore")]:
         ignore_errs = [FileNotFoundError, NotADirectoryError]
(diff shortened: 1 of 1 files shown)


alexreinking:
@henryiii -- I'm not understanding what failed in the CI runs or why. Can you take a look, please?

henryiii:
GHA was down: https://www.githubstatus.com

henryiii:
Thanks!

PR #895, opened 2024-09-10
fix: editable subpackage

henryiii:
Proposed fix for https://github.com/scikit-build/scikit-build-core/issues/893.

- **fix: nox + uv downstream improvements**
- **fix: editable subpackage issue**


src/scikit_build_core/resources/_editable_redirect.py
```diff
@@ -87,21 +87,26 @@ def find_spec(
...
+                else None,
             )
         if fullname in self.known_source_files:
             redir = self.known_source_files[fullname]
             return importlib.util.spec_from_file_location(
                 fullname,
                 redir,
-                submodule_search_locations=submodule_search_locations,
+                submodule_search_locations=submodule_search_locations
+                if redir.endswith("__init__.py")
```
henryiii, line 108:
```suggestion
                if redir.endswith(("__init__.py", "__init__.pyc"))
```

PR #916, opened 2024-09-27
fix: ignore build dir automatically

henryiii:
See #914.

### DIFF
--- a/src/scikit_build_core/build/_file_processor.py
+++ b/src/scikit_build_core/build/_file_processor.py
@@ -32,6 +32,7 @@ def each_unignored_file(
     starting_path: Path,
     include: Sequence[str] = (),
     exclude: Sequence[str] = (),
+    build_dir: str = "",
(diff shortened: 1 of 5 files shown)


src/scikit_build_core/build/_file_processor.py
```diff
@@ -50,9 +51,19 @@ def each_unignored_file(
...
 
+    exclude_build_dir = build_dir.format(
+        cache_tag="*",
+        wheel_tag="*",
+        build_type="*",
+        state="*",
+    )
+    exclude_lines = (
+        f"{EXCLUDE_LINES}\n{exclude_build_dir}" if exclude_build_dir else EXCLUDE_LINES
+    )
```
henryiii, line 62:
```suggestion
    
    exclude_lines = [*EXCLUDE_LINES, exclude_build_dir] if exclude_build_dir else EXCLUDE_LINES
```

PR #937, opened 2024-10-24
fix: correct the function prototype defined in the stub file in simplest_c example

LinZhihao-723:
In the simplest_c example, the stub file defines the function `square` to take two arguments. However, it only takes one in the C-extension implementation. This PR corrects the stub file to make it consistent with the C-extension implementation.


henryiii (approved):

henryiii:
Thanks!

PR #968, opened 2025-01-11
Lock during experimental editable.rebuild

hauntsaninja:


henryiii:
Can't find the run CI button for the new GitHub interface. Guess I'll switch back to approve, but that's odd that it seems to be missing...

henryiii:
I'm not sure using a dependency like this will work this way. Scikit-build isn't required to be installed when it rebuilds. So our dependency might not be there. That's why this file doesn't use any dependencies.

hauntsaninja:
I think that's a known issue with the new merge experience. 

Hmm, I could make something simple with fcntl and gate to Unix only? Or we could vendor?

henryiii:
Unix only would at least be better than the status quo. Vendoring would be hard since we don't even have scikit-build-core present in the user's environment after the editable install.

Pretty sure get_requires_for_build_editable is not what we need, we need to add dependencies to the dependencies itself. Maybe it's okay to do that, since editable installs aren't redistributable?

hauntsaninja:
Okay, I pushed a Unix only fcntl lock based on filelock's implementation.

henryiii:
Thanks!

PR #1004, opened 2025-02-27
ci: run some tests in parallel

henryiii:
A bit faster.


henryiii:
Great, this makes the filenames too long on Windows...

PR #1023, opened 2025-03-19
docs: fix incorrect configuration example

buddly27:
The documentation provides an incorrect example for disabling `site-packages` when searching for CMake configs. It should align with the value specified in the [schema](https://github.com/scikit-build/scikit-build-core/blob/main/src/scikit_build_core/resources/scikit-build.schema.json#L414)

### DIFF
--- a/docs/configuration/search_paths.md
+++ b/docs/configuration/search_paths.md
@@ -113,8 +113,8 @@ the `site-packages` folder where the project will be installed or the build
 isolation's `site-packages` folder. This default can be disabled by setting
 
 ```toml
-[tool.scikit-build.search]
-site-packages = false
+[tool.scikit-build]
+search.site-packages = false
 ```
 
 Additionally, scikit-build-core reads the entry-point `cmake.prefix` of the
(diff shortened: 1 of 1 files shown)


docs/configuration/search_paths.md
```diff
@@ -114,7 +114,7 @@ isolation's `site-packages` folder. This default can be disabled by setting
 
 ```toml
 [tool.scikit-build.search]
```
henryiii, line 116:
```suggestion
[tool.scikit-build]
```

docs/configuration/search_paths.md
```diff
@@ -114,7 +114,7 @@ isolation's `site-packages` folder. This default can be disabled by setting
...
-search.use-site-packages = false
+site-packages = false
```
henryiii, line 117:
```suggestion
search.site-packages = false
```

henryiii (commented):
Let's use this form, which used most other places.

LecrisUT (approved):
Yep, thanks for catching it.

PR #1048, opened 2025-04-18
fix: don't traverse into excluded directories

henryiii:
This avoids recursing into excluded directories, which fixes issues with adding
"README.md" to your include to find all excluded directories that have a readme
are also now included. There's now INFO logging that explains why things are
included or excluded.


jcfr:
It may be worth updating the documentation to describe that excludes are processed first, then includes.

PR #1103, opened 2025-06-18
Improve .gitignore iteration speed

silversquirl:
Using `os.walk` instead of `Path().rglob` can be 10x faster in deeply nested directory structures. This is a significant saving on certain projects, where just this iteration can take up to 3 or 4 seconds.

An even better approach would be to parse gitignore files lazily, as the directory tree is traversed later in this function. Another optimization would be to prune ignored directories while walking the directory tree, so they aren't even explored.

However, just the single optimization in this PR already reduces the impact of this code significantly, bringing it from almost 3s in my profile to just 0.4s.


silversquirl:
Can anyone confirm whether these CI failures are spurious? The docs one definitely is, but I'm not sure about the tests

PR #1126, opened 2025-08-11
tests: ignore setuptools-scm warning for packages that don't use it

henryiii:
Should fix our CI.


henryiii:
Yes, it's the non-isolated build test, and when it's present (for our own tests), it's now producing a warning on packages that don't use setuptools-scm at all.

henryiii:
(And it's a setuptools plugin, so it's imported by setuptools when processing entry points)

PR #1129, opened 2025-08-15
chore: add known wheels for riscv64

mayeut:
There are now cmake & ninja wheels available for riscv64, add them to the list of known wheels.

Side note for another issue/PR: should this list be updated to account for min/max versions at some point ? (e.g. don't request CMake 4.1.0 on manylinux1/2010 or musllinux_1_1)


henryiii:
> Side note for another issue/PR: should this list be updated to account for min/max versions at some point ? (e.g. don't request CMake 4.1.0 on manylinux1/2010 or musllinux_1_1)

I'd hope these are not being used (at least with scikit-build-core projects), and if they were, `cmake` is already present on those base images, so I think it would be pretty low impact, but I'd be open to it.

PR #1194, opened 2025-12-14
tests: consolidate editable and isolated fixtures

LecrisUT:
A bit of cleanup in preparation for #1193

### DIFF
--- a/tests/test_broken_fallback.py
+++ b/tests/test_broken_fallback.py
@@ -11,7 +11,7 @@ from scikit_build_core.build import (
 
 @pytest.mark.compile
 @pytest.mark.configure
-@pytest.mark.parametrize("package", {"broken_fallback"}, indirect=True)
+@pytest.mark.parametrize("package", ["broken_fallback"], indirect=True)
 @pytest.mark.usefixtures("package")
 @pytest.mark.parametrize("broken_define", ["BROKEN_CMAKE", "BROKEN_CODE"])
 def test_broken_code(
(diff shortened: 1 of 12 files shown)


LecrisUT:
Ahhh, the tests are failing at too long paths? Well that's really annoying

tests/conftest.py
```diff
@@ -192,158 +209,110 @@ def source_date_epoch(self) -> str:
...
-) -> PackageInfo:
-    package = PackageInfo(
-        "simple_purelib_package",
-    )
-    process_package(package, tmp_path, monkeypatch)
-    return package
-
+@dataclasses.dataclass(frozen=True)
+class Editable:
+    mode: str | None
```
eirrgang: Can we at least provide a hint of expected values with `typing.Literal`? Or a comment with a URL or something?

tests/conftest.py
```diff
@@ -192,158 +209,110 @@ def source_date_epoch(self) -> str:
...
+    pkg_src = DIR / "packages" / package.name
+    assert pkg_src.exists()
+    shutil.copytree(pkg_src, package.workdir, dirs_exist_ok=True)
+    monkeypatch.chdir(package.workdir)
 
 
 @pytest.fixture
-def package_simple_pyproject_source_dir(
-    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+def package(
```
eirrgang, line 227: I see. So this is now not just a factory for a `PackageInfo`, but takes care of staging the named package sources into a fresh temporary directory and changing into that directory, which imposes the normalized package layout that motivates all of the file moves. Test functions only need to capture the fixture as an argument if they want to check the hashes. I wish I could think of an alternative name to suggest. This is more readable, overall, I think. The parameterization is not intuitive, but the examples are clear.

tests/conftest.py
LecrisUT: Yeah sure thing :+1:

tests/conftest.py
LecrisUT: >  I wish I could think of an alternative name to suggest.

Always welcome if you have naming candidates.

> Test functions only need to capture the fixture as an argument if they want to check the hashes

Could expose some functionality for it to avoid needing to ignore the fixture in parameters, just haven't found a worthy one to do so with.

PR #1196, opened 2025-12-16
Respect _PYTHON_HOST_PLATFORM env variable

ryanking13:
Adds a check for `_PYTHON_HOST_PLATFORM` variable when producing the best wheel tag. I was guessing that this file should be the place I should put this, but please let me know if I am editing the wrong file.

Resolves: https://github.com/scikit-build/scikit-build-core/issues/920

### DIFF
--- a/docs/guide/crosscompile.md
+++ b/docs/guide/crosscompile.md
@@ -65,5 +65,5 @@ This is unfortunately incorrectly stripped from the cmake wrapper pyodide uses,
 so FindPython will report the wrong values, but pyodide-build will rename the
 .so's afterwards.
 
-pyodide-build will also set `_PYTHON_HOST_PLATFORM` to the target Pyodide platform,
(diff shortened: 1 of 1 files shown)


ryanking13:
> Looks generally ok, but do not have the context of how this is defined and by what, but assuming everything is fine on https://github.com/scikit-build/scikit-build-core/issues/920 side, I don't see a reason why not to add it. Thanks :)

I think this is an undocumented variable that is used by several packages, including sysconfig, cibuildwheel, auditwheel, setuptools ([link1](https://github.com/search?q=org%3Apypa%20_PYTHON_HOST_PLATFORM%20&type=code), [link2](https://github.com/python/cpython/blob/e27f76b6fe9340d9c92571a83e3e91baf52d140b/Lib/sysconfig/__init__.py#L121)). Originally used from distutils I guess.

ryanking13:
cc: @henryiii

PR #1199, opened 2025-12-22
tests: rework downloading the wheelhouse for isolation tests

henryiii:
Hopefully adding output capture will help us debug the intermittent failure on 3.12 • CMake 3.26 on windows-latest.

The problem is each worker is trying to download the same wheelhouse. I've reworked how the wheelhouse is setup, including a way to run the setup standalone so that something like Fedora could pre-download it before running pytest, no longer requiring the network. `python tests/utils/download_wheels.py` should do it if pytest is normally configured. We could put these in a top-level folder if that's better.

### DIFF
--- a/tests/test_hatchling.py
+++ b/tests/test_hatchling.py
@@ -3,9 +3,8 @@ import tarfile
 import zipfile
 from pathlib import Path
 
-import pytest
-
 import download_wheels
+import pytest
 
 pytest.importorskip("hatchling")
 
@@ -44,7 +43,9 @@ def test_hatchling_sdist(isolated, tmp_path: Path) -> None:
 )
 def test_hatchling_wheel(isolated, build_args, tmp_path: Path) -> None:
     dist = tmp_path / "dist"
-    isolated.install("build[virtualenv]", "scikit-build-core", "hatchling", *download_wheels.EXTRA)
+    isolated.install(
+        "build[virtualenv]", "scikit-build-core", "hatchling", *download_wheels.EXTRA
+    )
     isolated.module("build", "--no-isolation", f"--outdir={dist}", *build_args)
     ext_suffix = sysconfig.get_config_var("EXT_SUFFIX")
 
--- a/tests/utils/download_wheels.py
+++ b/tests/utils/download_wheels.py
@@ -13,7 +13,6 @@ import subprocess
 import sys
 from pathlib import Path
 
-
 EXTRA = []
 
 if importlib.util.find_spec("cmake") is not None:


.distro/python-scikit-build-core.spec
```diff
@@ -22,6 +22,7 @@ BuildRequires:  ninja-build
 BuildRequires:  gcc
 BuildRequires:  gcc-c++
 BuildRequires:  git
+BuildRequires:  python3-hatch-vcs
```
LecrisUT: No need to add it manually, it is already in the `-g test` of `%pyproject_buildrequires`

tests/conftest.py
```diff
@@ -36,55 +37,28 @@
 VIRTUALENV_VERSION = Version(metadata.version("virtualenv"))
 
 
-@pytest.fixture(scope="session")
-def pep518_wheelhouse(tmp_path_factory: pytest.TempPathFactory) -> Path:
-    wheelhouse = tmp_path_factory.mktemp("wheelhouse")
+def pytest_configure(config) -> None:
+    if hasattr(config, "workerinput"):
+        return  # Running in pytest-xdist worker
```
LecrisUT: Is it guaranteed to have run once without xdist though?

tests/conftest.py
```diff
@@ -36,55 +37,28 @@
...
+
+    wheelhouse = config.cache.mkdir("wheelhouse")
 
     subprocess.run(
         [
             sys.executable,
             "-m",
             "pip",
             "wheel",
+            "--no-build-isolation",
```
LecrisUT: Does running under build issolation or not make a difference here?

tests/conftest.py
```diff
@@ -36,55 +37,28 @@
...
-            "download",
-            "-q",
-            "-d",
-            str(wheelhouse),
-            *packages,
-        ],
-        check=True,
-    )
-    return wheelhouse
+    if not all(list(wheelhouse.glob(f"{p}*.whl")) for p in download_wheels.WHEELS):
```
LecrisUT: Btw, if you want to use `uv.lock` to guarantee some consistency and test more dependency drifts, feel free to do so. Just need an opt-in/out flag for building in the spec.

tests/conftest.py
```diff
@@ -36,55 +37,28 @@
 VIRTUALENV_VERSION = Version(metadata.version("virtualenv"))
 
 
-@pytest.fixture(scope="session")
-def pep518_wheelhouse(tmp_path_factory: pytest.TempPathFactory) -> Path:
-    wheelhouse = tmp_path_factory.mktemp("wheelhouse")
+def pytest_configure(config) -> None:
```
LecrisUT: Can you add an opt-out flag for this for building the rpm?

I checked again on how the building is done in fedora [^1] and still no dice on making these work there.

One hing we could do is make the venv construct with `--system-site-packages` and skip all tests with build-isolation, but we would need a different interface method for  installing dependencies that can be made no-op in that case.

[^1]: https://src.fedoraproject.org/rpms/pyproject-rpm-macros/blob/rawhide/f/pyproject_wheel.py#_43

.distro/python-scikit-build-core.spec
henryiii: I was trying to figure out why it wasn't working in the fedora jobs.

tests/conftest.py
henryiii: I believe so.

tests/conftest.py
henryiii: It requires internet if you run with build isolation. We could use the downloaded files, though, and keep build isolation.

tests/conftest.py
henryiii: If we can put the wheels in here somehow, it should work, but if fedora has no internet access, we'll need to disable it. I believe, looking at it now, that if you disabled all "network" tests, then the fixture was never computed, which is why it was working before. Maybe we should use a filelock on the fixture instead.

LecrisUT: > If we can put the wheels in here somehow, it should work, but if fedora has no internet access, we'll need to disable it

I need to push again about packaging the wheels, maybe even making a change proposal about it. But in the meantime disabling/auto-skipping the tests under there would be the most compatible approach right now. Having a pytest flag with `pytest_addoption` and skipping the wheelhouse setup should be enough in there since the tests are still being marked with `network` marker afaict

Edit: ah I see the file lock approach now

tests/conftest.py
```diff
@@ -37,53 +39,31 @@
...
-        "pip",
-        "setuptools",
-        "virtualenv",
-        "wheel",
-    ]
+def pep518_wheelhouse(pytestconfig: pytest.Config) -> Path:
+    wheelhouse = pytestconfig.cache.mkdir("wheelhouse")
+
+    main_lock = FileLock(wheelhouse / "main.lock")
+    with main_lock:
```
LecrisUT, line 46: I think you need to gate this by presence of scikit-build-core wheel.

Also don't think we need to 2 different locks?

henryiii: It's not an editable wheel, so it needs to be rebuilt every time the tests run - here it's being rebuilt by every process, but that's hopefully pretty cheap. Probably cheaper if we use build isolation.

henryiii: With two locks we can have one process start downloading wheels while the others are rebuilding the scikit-build-core wheel.

LecrisUT: > It's not an editable wheel, so it needs to be rebuilt every time the tests run

Wait I don't get this part. Isn't it building the wheel so that every subsequent `pip imstall` is taking that wheel? I am thinking about the guarding for across the workers, but maybe you are considering it across different commits? It should be checked across the `hatch-vcs` version, which I think you were trying to do previously?

> With two locks we can have one process start downloading wheels while the others are rebuilding the scikit-build-core wheel.

Good point, but that would only happen if it skips the locked functions gated by `main.lock` right? But speaking of which is it re-using the pip/uv cache? Using a `uv.lock` or equivalent could make the downloads be effectively no-ops

henryiii: It's testing the current source of scikit-build-core, so it has to be fresh. Yes, I'd like to only build it on one worker, but it's not that easy unless we make sure we always clean it up after a run (regardless of if the run failed, cancelled, etc).

LecrisUT: > It's testing the current source of scikit-build-core, so it has to be fresh. Yes, I'd like to only build it on one worker, but it's not that easy unless we make sure we always clean it up after a run (regardless of if the run failed, cancelled, etc).

I did some local debugging hoping that `hatch vcs` could help with the dirty files, but it seems like that would not work since it doesn't check the actual contents of the files when it is dirty, only the current date and if there are any dirty files.

The only alternative I can think of is to explicitly check against the `worker_id` and only run it on `master`, but are we able to do that without hard-coding `pytest-xdist` dependency?

LecrisUT:
The `testing-farm:fedora-rawhide-x86_64:upstream` is failing because of missing dependency in:
https://github.com/scikit-build/scikit-build-core/blob/501ff0f4a56e263c4dec61ed548cb0987ab61a12/tests/plans.fmf#L30

tests/conftest.py
LecrisUT: Am I reading the failure in https://github.com/scikit-build/scikit-build-core/actions/runs/20453587799/job/58771162110?pr=1199 correctly that using the `session` scope on multiple workers some runs went ahead as if the session fixture was already setup? It would be replicable if we artificially add a ridiculous timeout under the `wheels.lock`. Wonder which of the 2 approaches would be better.

PR #1200, opened 2025-12-23
tests: test-core, pybind11 in full test

henryiii:
Followup to #1188 and #1199. This adds `test-core`, for just the minimum required to run, and `test` is now a full test suite. `test-parallel` is now added.

### DIFF
--- a/.distro/python-scikit-build-core.spec
+++ b/.distro/python-scikit-build-core.spec
@@ -50,7 +50,7 @@ cp -p src/scikit_build_core/_vendor/pyproject_metadata/LICENSE LICENSE-pyproject
 
 %generate_buildrequires
 export HATCH_METADATA_CLASSIFIERS_NO_VERIFY=1
-%pyproject_buildrequires -g test-core%{?with_optional_tests:,test-parallel,test-meta,test-numpy,test-pybind11}
+%pyproject_buildrequires -g test-core%{?with_optional_tests:,test}
 
 
 %build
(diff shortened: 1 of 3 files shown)


.distro/python-scikit-build-core.spec
```diff
@@ -50,7 +50,7 @@ cp -p src/scikit_build_core/_vendor/pyproject_metadata/LICENSE LICENSE-pyproject
 
 %generate_buildrequires
 export HATCH_METADATA_CLASSIFIERS_NO_VERIFY=1
-%pyproject_buildrequires -g test%{?with_optional_tests:,test-meta,test-numpy,test-pybind11}
+%pyproject_buildrequires -g test-core%{?with_optional_tests:,test-parallel,test-meta,test-numpy,test-pybind11}
```
LecrisUT: Other than `test-schema` (which seems more fitting in `dev`), is there any group from `test` that should not be added?

.github/workflows/ci.yml
```diff
@@ -246,7 +246,9 @@ jobs:
             cmake ninja git make gcc-g++ python39 python39-devel python39-pip
 
       - name: Install
-        run: python3.9 -m pip install . --group=test-pybind11
+        run:
+          python3.9 -m pip install . --group=test-core --group=test-pybind11
+          --group=test-parallel --group=test-schema --group=test-meta
```
LecrisUT, line 252: Could at least document why this one has a special group set?

.distro/python-scikit-build-core.spec
henryiii: Moved it to dev.

.github/workflows/ci.yml
henryiii: Added a note (numpy doesn't have wheels).

PR #1206, opened 2026-01-09
chore: ignore lock files too

henryiii:
Getting a warning if one exists about it being placed in the SDist.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -173,3 +173,5 @@ Thumbs.db
 
 /docs/**/build
 .vscode/
+uv.lock
+*pylock


henryiii:
Depends on the frequency, but even weekly would be a lot of PRs with bumps and still hide breakage for a week.

LecrisUT:
> Depends on the frequency, but even weekly would be a lot of PRs with bumps and still hide breakage for a week.

True, but that is why I'm not a fan of dependabot because it closes and re-creates PRs. The workflow with renovatebot is more that you merge at your own pace, usually right before a release. It would still be noisy with rebases, so it's only valuable if it can notify of failing states.

Issue #1089, opened 2025-05-28
Ninja generator does not work on Windows due to (supposedly) scikit internal behavior


dograt:
Hi @henryiii . Is there a way to view and override the fact that `scikit-build-core` seems to be passing the architecture to CMake under the hood on Windows, even if I am using Ninja?

```
  CMake Error at CMakeLists.txt:16 (project):
    Generator

      Ninja

    does not support platform specification, but platform

      x64

    was specified.


  -- Configuring incomplete, errors occurred!
```

If it matters, I am using overrides to specify a CMake preset, which works fine on other OSs:
```
  if.platform-system = "^win32"
  build-dir = "build/desktop-windows"
  cmake.args = ["--preset=desktop-windows"]
```
I don't set the architecture anywhere myself, I so I'm assuming that `scikit-build-core` does this under the hood somewhere.

BLaZeKiLL:
observed this without presets also by just passing

```
cmake.args = ["-G Ninja"]
```

BLaZeKiLL:
> observed this without presets also by just passing
> 
> ```
> cmake.args = ["-G Ninja"]
> ```
> 
> I am using visual studio developer command shell, Ideally I want to use presets and compile using clang and directly invoking cmake works but I get the same error as reported when running scikit-build-core

First of all just `-G Ninja` failing was a cache issue but I looked further into presets

I looked into the code in `cmake.py, builder.py and generator.py` and found 2 things
```py
def get_generator(self, *args: str) -> str | None:
    """
    Try to get the generator that will be used to build the project. If it's
    not set, return None (default generator will be used).
    """
    generators = [g for g in args if g.startswith("-G")]
    if generators:
        return generators[-1][2:].strip()
    return self.env.get("CMAKE_GENERATOR", None)
```

In the even if your generator is specified in presets the detected generator is None as it only looks for `-G` so the following in `generator.py`, will default the generator to msvc on windows and set the platform then when cmake starts it will read the preset and update the generator but the platform is still passed as `x64` to the cache list

```py
def set_environment_for_gen(
    generator: str | None,
    cmake: CMake,
    env: MutableMapping[str, str],
    ninja_settings: NinjaSettings,
) -> Mapping[str, str]:
    """
    This function modifies the environment as needed to safely set a generator.
    You should have used CMAKE_GENERATOR already to get the input generator
    string.

    A reasonable default generator is set if the environment does not already
[...]

PR #1209, opened 2026-01-21
fix: better support for ninja on Windows

henryiii:
Fix for #1089. Two fixes:

* The first is using `CMAKE_VS_PLATFORM_NAME_DEFAULT` instead on Windows. I'm not sure how well that will work, but can try it.
* The second supports using define instead of just args when checking for a set
  generator. That won't help preset usage, unless you set it here too, but
  that's easier than setting it via args.

### DIFF
--- a/src/scikit_build_core/builder/generator.py
+++ b/src/scikit_build_core/builder/generator.py
@@ -101,10 +101,8 @@ def set_environment_for_gen(
             logger.debug("Default generator: {}", generator)
 
     if sysconfig.get_platform().startswith("win") and "Visual Studio" in generator:
-        # This must also be set when *_PLATFORM is set.
-        env.setdefault("CMAKE_GENERATOR", generator)
-        env.setdefault("CMAKE_GENERATOR_PLATFORM", get_cmake_platform(env))
-        return {}
+        # This ensures we default to the correct platform for Python, can be overridden
+        return {"CMAKE_VS_PLATFORM_NAME_DEFAULT": get_cmake_platform(env)}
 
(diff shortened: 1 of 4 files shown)


src/scikit_build_core/builder/generator.py
```diff
@@ -101,10 +101,8 @@ def set_environment_for_gen(
             logger.debug("Default generator: {}", generator)
 
     if sysconfig.get_platform().startswith("win") and "Visual Studio" in generator:
-        # This must also be set when *_PLATFORM is set.
-        env.setdefault("CMAKE_GENERATOR", generator)
-        env.setdefault("CMAKE_GENERATOR_PLATFORM", get_cmake_platform(env))
-        return {}
+        # This ensures we default to the correct platform for Python, can be overridden
+        return {"CMAKE_VS_PLATFORM_NAME_DEFAULT": get_cmake_platform(env)}
```
LecrisUT, line 105: Could make this independet of generator to support the opposite (defining the VS generator in the preset), should not affect the other generators iiuc.

henryiii: Then it shouldn't return early, not sure if the latter part (setting CXX and such) would mess up MSVC.

LecrisUT: Hmm, good point. I think it would be a common issue for all generators, but windows setup might be more special in if the `CC` in the `sysconfig` is actually relevant. I guess the current check would still have the least impact for now until we can figure out if we want to relax it.

I guess mingw could serve as a test environment for that? Not sure how all of these windows toolchains work.

PR #1214, opened 2026-02-01
Support PEP 604 unions in sources (#1213)

### LINKED ISSUE #1213: Support PEP 604 Unions
`scikit-build-core` expects unions to be of type `typing.Union`.  This is problematic if a project based on `scikit-build-core` such as [`scikit-build-core-conan`](https://github.com/wu-vincent/scikit-build-core-conan)) uses [PEP 604](https://peps.python.org/pep-0604/) unions which are of type `types.UnionType`.  See wu-vincent/scikit-build-core-conan#11

I believe the fix is to simply check for both `typing.Union` or `types.UnionType` whenever checking for a union.

martinezlc99:
Adds support for PEP 604 unions.

Closes #1213

### DIFF
--- a/src/scikit_build_core/settings/sources.py
+++ b/src/scikit_build_core/settings/sources.py
@@ -49,6 +49,7 @@ from __future__ import annotations
 
 import dataclasses
 import os
+import sys
 import typing
 from typing import Any, Literal, Protocol, TypeVar, Union
 
@@ -57,6 +58,16 @@ from .._compat.typing import Annotated, get_args, get_origin
 
 if typing.TYPE_CHECKING:
     from collections.abc import Generator, Iterator, Mapping, Sequence
+
+# Runtime check for PEP 604 union syntax (A | B) support
+# types.UnionType only exists in Python 3.10+
+if sys.version_info >= (3, 10):
+    from types import UnionType as _UnionType  # type: ignore[attr-defined]
+else:
+
+    class _UnionType:
+        pass
+
 
 T = TypeVar("T")
 
@@ -94,14 +105,14 @@ def _dig_fields(opt: Any, /, *names: str) -> Any:
     return opt
 
 
-def _process_union(target: type[Any], /) -> Any:
+def _process_union(target: Any, /) -> Any:
     """
     Filters None out of Unions. If a Union only has one item, return that item.
     """
 
     origin = get_origin(target)
 
-    if origin is Union:
+    if _is_union_type(origin):
         non_none_args = [a for a in get_args(target) if a is not type(None)]
         if len(non_none_args) == 1:
             return non_none_args[0]
@@ -133,6 +144,14 @@ def _get_target_raw_type(target: type[Any] | Any, /) -> Any:
     target = _process_union(target)
     origin = get_origin(target)
     return origin or target
+
+
+def _is_union_type(raw_target: Any) -> bool:
+    """
+    Check if raw_target is a Union type (either ``typing.Union`` or ``types.UnionType``).
+    Handles both ``typing.Union[A, B]`` and PEP 604 syntax (``A | B``).
+    """
+    return raw_target is Union or raw_target is _UnionType
 
 
 def _get_inner_type(_target: type[Any], /) -> type[Any]:
@@ -268,10 +287,10 @@ class EnvSource:
         if raw_target is bool:
             return _process_bool(item)
 
-        if raw_target is Union and str in get_args(target):
+        if _is_union_type(raw_target) and str in get_args(target):
             return item
 
-        if raw_target is Union:
+        if _is_union_type(raw_target):
             args = {_get_target_raw_type(t): t for t in get_args(target)}
             if str in args:
                 return item
@@ -407,7 +426,7 @@ class ConfSource:
         if raw_target is dict:
             assert not isinstance(item, (str, list, bool))
             return {k: cls.convert(v, _get_inner_type(target)) for k, v in item.items()}
-        if raw_target is Union:
+        if _is_union_type(raw_target):
             args = {_get_target_raw_type(t): t for t in get_args(target)}
             if str in args:
                 return item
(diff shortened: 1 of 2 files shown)


src/scikit_build_core/settings/sources.py
```diff
@@ -94,14 +102,18 @@ def _dig_fields(opt: Any, /, *names: str) -> Any:
...
+def _process_union(target: Any, /) -> Any:
     """
     Filters None out of Unions. If a Union only has one item, return that item.
     """
 
     origin = get_origin(target)
 
-    if origin is Union:
+    # Handle PEP 604 union syntax which has origin types.UnionType
+    if origin is _UnionType:
```
henryiii: If origin is None, this could be true on python 3.9 and 3.8? Is there a reason you don’t have to protect it with a check on _UnionType being None? Is origin and raw_target always non None?

src/scikit_build_core/settings/sources.py
```diff
@@ -58,6 +59,13 @@
 if typing.TYPE_CHECKING:
     from collections.abc import Generator, Iterator, Mapping, Sequence
 
+# Runtime check for PEP 604 union syntax (A | B) support
+# types.UnionType only exists in Python 3.10+
+if sys.version_info >= (3, 10):
+    from types import UnionType as _UnionType  # type: ignore[attr-defined]
+else:
+    _UnionType = None
```
LecrisUT: Instead of linking this to `None`, define an arbitrary class which would never validate.

src/scikit_build_core/settings/sources.py
```diff
@@ -94,14 +102,18 @@ def _dig_fields(opt: Any, /, *names: str) -> Any:
...
     """
     Filters None out of Unions. If a Union only has one item, return that item.
     """
 
     origin = get_origin(target)
 
-    if origin is Union:
+    # Handle PEP 604 union syntax which has origin types.UnionType
+    if origin is _UnionType:
+        origin = Union
```
LecrisUT: Why is `origin` being assigned here? Shouldn't the `_is_union_type` catch it?

tests/test_sources_pep604_union.py
```diff
@@ -0,0 +1,36 @@
+# mypy: ignore-errors
+import sys
+from typing import Union, get_origin
+
+import pytest
+
+# Skip entire module on Python < 3.10
+pytestmark = pytest.mark.skipif(
+    sys.version_info < (3, 10), reason="PEP 604 union syntax requires Python 3.10+"
```
LecrisUT: We should still test that it works on lower python versions. Can you adjust it to conditionally test both?

src/scikit_build_core/settings/sources.py
martinezlc99: Yes, you are correct.  I will update.

src/scikit_build_core/settings/sources.py
```diff
@@ -94,14 +105,14 @@ def _dig_fields(opt: Any, /, *names: str) -> Any:
     return opt
 
 
-def _process_union(target: type[Any], /) -> Any:
+def _process_union(target: Any, /) -> Any:
```
LecrisUT, line 108: What is the purpose of this change?

src/scikit_build_core/settings/sources.py
martinezlc99: Yes, unnecessary. Removed

src/scikit_build_core/settings/sources.py
martinezlc99: Honestly, it was to silence a static analysis error in the added test since types of most annotations themselves are not types.  I did notice some methods in `sources.py` had both, so I updated.  Since I don't believe these are actual runtime errors, I can change back to `type[Any] ` although mypy may complain in the test.

I also didn't notice other tests were not class based so depending on your answer to the above I can switch to method based tests as well...or can switch to method based tests regardless.

henryiii: The types on these are pretty bad. Python typing doesn't have good meta typing; [PEP 0747](https://peps.python.org/pep-0747/) is needed to make this better, I think. Until then it's a bit fuzzy.

henryiii: (that is, the support for types on things like this is bad)

henryiii:
FWIW, this distinction [went away in 3.14](https://docs.python.org/3/whatsnew/3.14.html#typing), but this is needed for 3.10-3.13.

src/scikit_build_core/settings/sources.py
LecrisUT: > Python typing doesn't have good meta typing; [PEP 0747](https://peps.python.org/pep-0747/) is needed to make this better, I think.

How would it look with PEP747? From what I tried the only improvement was in generics, but even those can be replicated if you are really really careful on how you write them.

Also @henryiii, comments on

> I also didn't notice other tests were not class based

Yay or nay for class based tests? I am indiferrent, and if anything would prefer the unit tests to reflect the src structure.

henryiii: Special typing forms are not `type[Any]`, they don't really have a type. This would give us a way to include these, I believe, since we need to handle several of them.

henryiii: I'm indifferent to them being class-based in this file. It feels like extra indentation and useless selfs just for a tiny bit of origination in a small file. If it was larger, I think more files or directories is better than classes when you aren't using `self`.

henryiii (approved):

src/scikit_build_core/settings/sources.py
LecrisUT: > Special typing forms are not `type[Any]`, they don't really have a type. This would give us a way to include these, I believe, since we need to handle several of them.

Ah ok, I think I understand, and dang the typing anotation for that is hard. Technically we can define it with overloads, but I don't even know if mypy can parse it (don't think it can handle `type[Union[T]]`. Sounds fine enough to deffer

PR #1218, opened 2026-02-16
docs: color CLI on Python 3.14+

henryiii:
Adding color to the CLI docs.

### DIFF
--- a/docs/ext/progout.py
+++ b/docs/ext/progout.py
@@ -31,6 +31,9 @@ class ShowCliDirective(Directive):
         # Get the cwd option, defaulting to the current file's directory
         cwd_option = self.options.get("cwd")
         cwd = (current_dir / cwd_option).resolve() if cwd_option else current_dir
+        env = os.environ
+        env["FORCE_COLOR"] = "1"
+        env.pop("NO_COLOR", None)
 
         try:
             # Run the command and capture output
@@ -40,10 +43,10 @@ class ShowCliDirective(Directive):
                 check=True,
                 text=True,
                 cwd=str(cwd),
-                env={**os.environ, "FORCE_COLOR": "1"},
+                env=env,
             )
             output = result.stdout or result.stderr
-        except Exception as e:  # noqa: BLE001
+        except subprocess.CalledProcessError as e:
             return [
                 nodes.error(
                     None,


docs/ext/progout.py
```diff
@@ -0,0 +1,160 @@
...
+
+        try:
+            # Run the command and capture output
+            result = subprocess.run(
+                command.split(),
+                capture_output=True,
+                check=True,
+                text=True,
+                cwd=str(cwd),
+                env={**os.environ, "FORCE_COLOR": "1"},
```
LecrisUT: Iiuc this is the only trick used to get the ansi codes to pop out right? Otherwise it seems that the upstream project had the plans to support the ansi  formatting. Maybe we can check what they are lacking to support this.

henryiii: Correct. But you need to transform the ANSI codes, which is what the upstream project is lacking, and I don't know if there's interest unless there's a project to do this. Rich can do it but it's a little hacky, and plubmum is close to being able to do it, I could probably improve support there.

LecrisUT: Do any of them allow for some palette configuration?

henryiii: I tried with rich. It works, but the colors aren't as nice in dark mode, and some of them don't register for some reason, like both the colors for the first box.


```diff
@@ -127,12 +63,20 @@ class ShowCliDirective(Directive):
             ]

         # Convert ANSI codes to HTML
-        html_output = AnsiToHtmlConverter().convert(output)
+        rich_text = rich.text.Text.from_ansi(output)
+
+        console = rich.console.Console(record=True, force_terminal=True)

         # Add the run block if this was `command-output`
         if self.name == "command-output":
-            color = ANSI_COLORS["90"]
-            html_output = f'<span style="color: {color};">$</span> <span style="font-weight: bold;">{command}</span>\n{html_output}'
+            console.print(f"[grey]$[/grey] [bold]{command}")
+
+        console.print(rich_text)
+        page = console.export_html(inline_styles=True)
+        result = re.search(r"<body.*?>(.*?)</body>", page, re.DOTALL | re.IGNORECASE)
+        assert result
+        txt = result.group(1).strip()
+        html_output = txt.replace("code ", 'code class="nohighlight" ')

         # Create a raw HTML node with the colored output
         raw_html = nodes.raw("", html_output, format="html")
```

LecrisUT: Handling the html sounds like anything could go wrong :shrug:

henryiii: Using https://github.com/tomerfiliba/plumbum/pull/774:

```diff
@@ -127,12 +55,11 @@ class ShowCliDirective(Directive):
             ]

         # Convert ANSI codes to HTML
-        html_output = AnsiToHtmlConverter().convert(output)
+        html_output = htmlcolors.sequence_to_string(htmlcolors.from_ansi_string(output))

         # Add the run block if this was `command-output`
         if self.name == "command-output":
-            color = ANSI_COLORS["90"]
-            html_output = f'<span style="color: {color};">$</span> <span style="font-weight: bold;">{command}</span>\n{html_output}'
+            html_output = htmlcolors.fg.light_gray.wrap('$') + " " + htmlcolors.bold.wrap(command) + "\n" + html_output

         # Create a raw HTML node with the colored output
         raw_html = nodes.raw("", html_output, format="html")
```

(and all custom ANSI code removed)

henryiii: The plumbum output isn't as nicely themed, though, the colors don't work in dark mode as well. Might be able to add support for theming, though.

https://github.com/erbsland-dev/erbsland-sphinx-ansi by @erbsland-dev is now out, but I don't know if that supports being used as a library[^1], and it's Python 3.12+[^2], which is a bit restrictive (we still are 3.8+, I could see 3.10+ (EoL), but 3.12+ is a bit much. Though we don't have to allow docs to be built with older Pythons, uv will refuse to resolve and needs workarounds to handle such a dependency. (Same thing would happen with pypa/packaging, also 3.8+ still.)

[^1]: I'm not all that familiar with sphinx, maybe one extension can call another? Actually, it does look like they can, though there are some choices to be made on if internals should be used or rst be reprocessed.

[^2]: It's not statically typed, which AFAIK is the main reason to go 3.12+.

I'd like to see the program output extension gain support for the ansi extension, that would be the ideal case.

erbsland-dev: @henryiii For [erbsland-sphinx-ansi](https://github.com/erbsland-dev/erbsland-sphinx-ansi), I’ve updated the Python compatibility to 3.10+ and introduced a new `theme` parameter that enables straightforward theming via custom CSS files.

henryiii: Excellent, I've updated to use it. I've made it 3.10+ only, and if it's missing, we strip ansi codes.

PR #1220, opened 2026-02-19
docs: add buildgen to list in quickstart

### LINKED ISSUE #1207: buildgen: a tool for generating scikit-build-core recipes
[buildgen](https://github.com/shakfu/buildgen) is a build system generator package supporting Makefile, CMake, and scikit-build-core project definitions.

Basically, I wrote this tool because I kept wanting to use `scikit-build-core` with `uv` when building python extensions and a cmake backend and makefile frontend for c/c++ projects.

It is pip installable via `pip install buildgen` ([pypi-project](https://pypi.org/project/buildgen/)), and it allows you to do the following:

```sh
# Create a C++ project with cmake build-backend and Makefile frontend.
buildgen new myapp

# Create a Python extension with pybind11 using scikit-build-core and uv
buildgen new myext -r py/pybind11

# List available recipes
buildgen list
```

It has the following features:

- **Makefile Generation**: Programmatic Makefile creation with variables, pattern rules, conditionals
- **CMake Generation**: Full CMakeLists.txt generation with find_package, FetchContent, install rules
- **Cross-Generator**: Define project recipes once in JSON/YAML, generate build systems
- **CMake Frontend**: Use CMake as build system with convenient Makefile frontend
- **Project Templates**: Quick-start templates for common project types
- **scikit-build-core Templates**: Python extension project scaffolding (pybind11, cython, nanobind, C)
- **Template Customization**: Override templates per-project, per-user, or via environment variable (Mako syntax)
- **Configurable Project Recipes**: 2-step JSON/YAML recipes which include options and which are `rendered` to generate the project infrastructure.
- **Full recipe testing**: all recipes are generated, built and tested using pytest.

henryiii:
Close #1207.

### DIFF
--- a/docs/guide/getting_started.md
+++ b/docs/guide/getting_started.md
@@ -6,6 +6,16 @@ package in pure Python using modern tooling and configuration. Another great
 resource is the [Scientific Python Developer Guide][]. And a tutorial can be
 found at [INTERSECT Training: Packaging][].
 
+You can also find examples:
+
+- For pybind11, there's a example template at [pybind11/scikit_build_example][].
+  For nanobind, [nanobind example][] includes the Stable ABI on Python 3.12+!
+- There are several examples including scikit-build-core examples (including
+  free-threading) at [scikit-build-sample-projects][].
+
+We also keep
+[a list of some of the projects using scikit-build-core](../about/projects.md).
+
 ## Quick start
 
 There are several mechanisms to quickly get started with a package:
@@ -18,12 +28,8 @@ There are several mechanisms to quickly get started with a package:
   `buildgen new myext -r py/pybind11` (see all with `buildgen list`, includes
   pybind11, nanobind, Cython, and C extensions).
 
-You can also find examples:
-
-- For pybind11, there's a example template at [pybind11/scikit_build_example][].
-  For nanobind, [nanobind example][] includes the Stable ABI on Python 3.12+!
-- There are several examples including scikit-build-core examples (including
-  free-threading) at [scikit-build-sample-projects][].
+For the rest of this page, however, we will show you how to get set up from
+scratch.
 
 ## Writing an extension
 


docs/guide/getting_started.md
```diff
@@ -423,12 +429,13 @@ $ pip install .
 That's it for a basic package!
 
 <!-- prettier-ignore-start -->
+[INTERSECT Training: Packaging]:     https://intersect-training.org/packaging
```
LecrisUT, line 438: Are we prettier ignoring these or not? 😅

docs/guide/getting_started.md
```diff
@@ -14,6 +14,12 @@ There are several mechanisms to quickly get started with a package:
   your package and run: `uv init --lib --build-backend=scikit`.
 - [scientific-python/cookie][] has a cookiecutter/copier template for making a
   package with all the suggestions in the [Scientific Python Developer Guide][].
+- [buildgen][] can generate Python extensions with
+  `buildgen new myext -r py/pybind11` (see all with `buildgen list`, includes
+  pybind11, nanobind, Cython, and C extensions).
+
+You can also find examples:
```
LecrisUT: "Or you can check out example ptojects"?

docs/guide/getting_started.md
henryiii: I manually sorted them. ;)

docs/guide/getting_started.md
henryiii: To me that sounds like a list of ptojects, that is, basically what we have https://scikit-build-core.readthedocs.io/en/latest/about/projects.html. The ones here are only examples, not full ptojects.

LecrisUT: Yeah, words are hard. Meant more about the flow in relation to the previous part (if we get rid of the list, does the paragraph still flow well)

henryiii: Moved it a little, and cross-linked. I think that's a bit better.

PR #1224, opened 2026-02-20
docs: source link

henryiii:
Adds a link back to the source repo, and fixes a WARNING admonition that slips through into the docs without formatting (see https://github.com/executablebooks/MyST-Parser/issues/845).


LecrisUT (approved):
> > Where is that?
>
> It's in the ToC navigation.

Ah I was blind, for some reason I thought the change was in some myst page metadata.

Issue #1197, opened 2025-12-17
Permissions get dropped when building wheel


jranieri-grammatech:
We ran into an issue where file permissions for package data were getting dropped when creating the wheel, even though they were correct in the git repo and the generated sdist. This appears to be due to [using `shutil.copyfile` instead `shutil.copy` when copying the file to the temporary directory](https://github.com/scikit-build/scikit-build-core/blob/20153a10c0e967198cd1a47b4b0796a6aae5d6d0/src/scikit_build_core/build/wheel.py#L484):
```python
        if not editable:
            for filepath, package_dir in mapping.items():
                Path(package_dir).parent.mkdir(exist_ok=True, parents=True)
                shutil.copyfile(filepath, package_dir)
```

I tested locally that changing it to use `shutil.copy` fixed our problem. Is this intentional?

PR #1225, opened 2026-02-20
fix: use copy2 instead of copyfile for wheels

henryiii:
Seeing what fixing #1197 might affect.

### DIFF
--- a/src/scikit_build_core/build/wheel.py
+++ b/src/scikit_build_core/build/wheel.py
@@ -354,7 +354,7 @@ def _build_wheel_impl_impl(
         for x in license_paths:
             path = wheel_dirs["metadata"] / "licenses" / x
             path.parent.mkdir(parents=True, exist_ok=True)
-            shutil.copy(x, path)
+            shutil.copy2(x, path)
 
         if (
             settings.wheel.license_files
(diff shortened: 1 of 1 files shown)


src/scikit_build_core/build/wheel.py
```diff
@@ -488,7 +488,7 @@ def _build_wheel_impl_impl(
         if not editable:
             for filepath, package_dir in mapping.items():
                 Path(package_dir).parent.mkdir(exist_ok=True, parents=True)
-                shutil.copyfile(filepath, package_dir)
+                shutil.copy(filepath, package_dir)
```
LecrisUT:
```suggestion
                shutil.copy2(filepath, package_dir)
```
Let's actually try to go with `copy2`?

> [copy()](https://docs.python.org/3/library/shutil.html#shutil.copy) copies the file data and the file’s permission mode (see [os.chmod()](https://docs.python.org/3/library/os.html#os.chmod)). Other metadata, like the file’s creation and modification times, is not preserved. To preserve all file metadata from the original, use [copy2()](https://docs.python.org/3/library/shutil.html#shutil.copy2) instead.

There is another occurrence of `copy` that we can change to `copy2`

PR #1230, opened 2026-02-23
chore: bump pyproject-metadata vendor to 0.11

henryiii:
Bumping this.


LecrisUT:
One more reference to bump it at:
https://github.com/scikit-build/scikit-build-core/blob/f4429b08316a485835c57e0beff256a7b505bc6c/.distro/python-scikit-build-core.spec#L38

PR #1238, opened 2026-02-26
feat: adding force include

henryiii:
First draft of force-include, with the help of copilot in VSCode.

### DIFF
--- a/src/scikit_build_core/settings/skbuild_model.py
+++ b/src/scikit_build_core/settings/skbuild_model.py
@@ -493,6 +493,30 @@ class GenerateSettings:
 
 
 @dataclasses.dataclass
+class ForceIncludeSettings:
+    source: Path
+    """
+    Source path to include.
+
+    Can be a file or a directory, and may point outside the project root.
+    """
+
+    sdist: str = ""
+    """
+    Destination path inside the SDist.
+
+    Leave empty to skip adding this source to the SDist.
+    """
+
+    wheel: str = ""
+    """
+    Destination path inside the wheel's purelib/platlib area.
+
+    Leave empty to skip adding this source to the wheel.
+    """
+
+
+@dataclasses.dataclass
 class MessagesSettings:
     """
     Settings for messages.
@@ -520,6 +544,7 @@ class ScikitBuildSettings:
     editable: EditableSettings = dataclasses.field(default_factory=EditableSettings)
     build: BuildSettings = dataclasses.field(default_factory=BuildSettings)
     install: InstallSettings = dataclasses.field(default_factory=InstallSettings)
+    force_include: List[ForceIncludeSettings] = dataclasses.field(default_factory=list)
     generate: List[GenerateSettings] = dataclasses.field(default_factory=list)
     messages: MessagesSettings = dataclasses.field(default_factory=MessagesSettings)
     search: SearchSettings = dataclasses.field(default_factory=SearchSettings)
--- a/README.md
+++ b/README.md
@@ -259,6 +259,15 @@ install.components = []
 # Whether to strip the binaries.
 install.strip = true
 
+# Source path to include.
+force-include[].source = ""
+
+# Destination path inside the SDist.
+force-include[].sdist = ""
+
+# Destination path inside the wheel's purelib/platlib area.
+force-include[].wheel = ""
+
 # The path (relative to platlib) for the file to generate.
 generate[].path = ""
 
--- a/docs/reference/configs.md
+++ b/docs/reference/configs.md
@@ -291,6 +291,35 @@ print(mk_skbuild_docs())
   Turn on verbose output for the editable mode rebuilds.
 ```
 
+## force-include[]
+
+```{eval-rst}
+.. confval:: force-include[].sdist
+  :type: ``str``
+
+  Destination path inside the SDist.
+
+  Leave empty to skip adding this source to the SDist.
+```
+
+```{eval-rst}
+.. confval:: force-include[].source
+  :type: ``Path``
+
+  Source path to include.
+
+  Can be a file or a directory, and may point outside the project root.
+```
+
+```{eval-rst}
+.. confval:: force-include[].wheel
+  :type: ``str``
+
+  Destination path inside the wheel's purelib/platlib area.
+
+  Leave empty to skip adding this source to the wheel.
+```
+
 ## generate[]
 
 ```{eval-rst}
--- a/src/scikit_build_core/build/_force_include.py
+++ b/src/scikit_build_core/build/_force_include.py
@@ -1,0 +1,36 @@
+from __future__ import annotations
+
+import shutil
+import typing
+
+if typing.TYPE_CHECKING:
+    from collections.abc import Iterator
+    from pathlib import Path
+
+
+__all__ = ["copy_force_include", "force_include_files"]
+
+
+def __dir__() -> list[str]:
+    return __all__
+
+
+def force_include_files(source: Path, destination: Path) -> Iterator[tuple[Path, Path]]:
+    if source.is_file():
+        yield source, destination
+        return
+
+    if source.is_dir():
+        for source_file in source.rglob("*"):
+            if source_file.is_file():
+                yield source_file, destination / source_file.relative_to(source)
+        return
+
+    msg = f"force-include source does not exist or is not a file/directory: {source}"
+    raise FileNotFoundError(msg)
+
+
+def copy_force_include(source: Path, destination: Path) -> None:
+    for source_file, destination_file in force_include_files(source, destination):
+        destination_file.parent.mkdir(parents=True, exist_ok=True)
+        shutil.copy2(source_file, destination_file)
--- a/src/scikit_build_core/build/sdist.py
+++ b/src/scikit_build_core/build/sdist.py
@@ -17,6 +17,7 @@ from .._compat import tomllib
 from .._logging import rich_print
 from ..settings.skbuild_read_settings import SettingsReader
 from ._file_processor import each_unignored_file
+from ._force_include import force_include_files
 from ._init import setup_logging
 from .generate import generate_file_contents
 from .metadata import get_standard_metadata
@@ -165,6 +166,21 @@ def build_sdist(
                 arcname=srcdirname / filepath,
                 filter=normalize_tar_info if reproducible else lambda x: x,
             )
+
+        for force_include in settings.force_include:
+            if not force_include.sdist:
+                continue
+
+            source = force_include.source
+            destination = Path(force_include.sdist)
+            for source_file, destination_file in force_include_files(
+                source, destination
+            ):
+                tar.add(
+                    source_file,
+                    arcname=srcdirname / destination_file,
+                    filter=normalize_tar_info if reproducible else lambda x: x,
+                )
 
         add_bytes_to_tar(
             tar, pkg_info, f"{srcdirname}/PKG-INFO", normalize=reproducible
--- a/src/scikit_build_core/build/wheel.py
+++ b/src/scikit_build_core/build/wheel.py
@@ -25,6 +25,7 @@ from ..errors import FailedLiveProcessError
 from ..format import pyproject_format
 from ..settings.skbuild_read_settings import SettingsReader
 from ._editable import editable_redirect, libdir_to_installed, mapping_to_modules
+from ._force_include import copy_force_include
 from ._init import setup_logging
 from ._pathutil import (
     packages_to_file_mapping,
@@ -489,6 +490,15 @@ def _build_wheel_impl_impl(
             for filepath, package_dir in mapping.items():
                 Path(package_dir).parent.mkdir(exist_ok=True, parents=True)
                 shutil.copy2(filepath, package_dir)
+
+            for force_include in settings.force_include:
+                if not force_include.wheel:
+                    continue
+
+                copy_force_include(
+                    force_include.source,
+                    wheel_dirs[targetlib] / force_include.wheel,
+                )
 
             process_script_dir(wheel_dirs["scripts"])
 
--- a/src/scikit_build_core/resources/scikit-build.schema.json
+++ b/src/scikit_build_core/resources/scikit-build.schema.json
@@ -352,6 +352,32 @@
         }
       }
     },
+    "force-include": {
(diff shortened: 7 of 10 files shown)


src/scikit_build_core/settings/skbuild_model.py
```diff
@@ -492,6 +492,30 @@ class GenerateSettings:
...
+
+    Can be a file or a directory, and may point outside the project root.
+    """
+
+    sdist: str = ""
+    """
+    Destination path inside the SDist.
+
+    Leave empty to skip adding this source to the SDist.
+    """
```
LecrisUT, line 509: Having each item like this just to support writing to either sdist or wheel seems messy for more items that are added. Hatch's format seems better.

We could either flip the order like in hatch and make them `force-inclued.(sdist|wheel)` each being a dict, or we keep just the one key `force-include` and if they want to specify differently for sdist or wheel then we have the overrides mechanism that can already do that.

henryiii:
Trying an alternative. Here's the plan (after about 4 iterations):

<details><summary>:robot: AI Plan :robot:</summary>
<p>

# Implement top-level `force-include` (sdist / wheel / build targets)

## Context

scikit-build-core can only place files in a distribution via package
auto-discovery (`wheel.packages`), gitignore-style `sdist.include`, or by
*generating* them (`generate[]`). There is no way to map an arbitrary existing
file or directory — especially one built elsewhere or living outside the
package tree — to a chosen destination, the way hatchling's `force-include`
table does. This adds that, with a scikit-build-core twist: a single top-level
table whose values can independently target the sdist, the wheel, and the CMake
build directory.

### Config shape (decided with the user)

A **top-level** `tool.scikit-build.force-include` table. Keys are source paths;
values are either a bare string or an inline table:

```toml
[tool.scikit-build.force-include]
# bare string -> WHEEL ONLY (leading-slash selects a wheel tree, experimental)
"vendor/lib.so"   = "pkg/_lib.so"
"tools/run.sh"    = "/scripts/run.sh"
# inline table -> per-target destinations, any subset of {sdist, wheel, build}
"../shared/data"  = {sdist = "data", wheel = "pkg/data"}
"toolchain.cmake" = {build = "toolchain.cmake"}
```

**Semantics:**
- key = source (project-root-relative; may be outside the tree, e.g. `../x`;
  `~` expanded). May be a **file or a directory** (directories recurse).
- bare string → wheel destination only.
- inline table → independent `sdist` / `wheel` / `build` destinations; omit a
  key to skip that target.
- **wheel** dest is platlib-relative; a leading `/data`, `/scripts`,
  `/headers`, `/platlib`, `/metadata` targets that wheel tree (EXPERIMENTAL,
  same gate + `wheel_dirs` validation as `wheel.install-dir`).
- **sdist** dest is relative to the sdist root.
- **build** dest is relative to the CMake build directory; copied **before
  configure** so CMake can consume it (only when CMake runs).
- directory recursion skips built-in junk (`__pycache__/`, `*.pyc`, `.git`,
  `.egg-info/`, …) by reusing `EXCLUDE_LINES` from `_file_processor.py`.
- force-include is *forced*: wheel files are copied **last**, overriding package
  files and CMake output at the same destination.
- **Missing source → skipped, not an error** (matches hatchling). Building from
  an sdist may lack an external (`../`) source, but it is expected to already be
  at the destination; we can't reliably distinguish a source-tree build from an
  sdist build, so stay lenient and `logger.debug`.

## Feasibility note (already investigated)

`Dict[str, str]` is supported today. The inline-table form needs
`Dict[str, Union[str, ForceIncludeTargets]]`, which the settings machinery does
**not** yet support — contained additions required (see step 2):
- `json_schema.py` already emits the correct `oneOf` — no change.
- `sources.py` `TOMLSource.convert()` union branch rejects a `dict` value today
  (it only matches when `type(item)` is exactly a union member, ~line 726-738).
- `documentation.py` doesn't render a non-Optional `Union` and doesn't expand a
  dataclass nested inside a dict value.
- `EnvSource`/`ConfSource` forbid dataclasses-in-dicts, so the **inline-table
  form is TOML-only**. Acceptable: force-include keys are file paths (dots,
  slashes) that don't round-trip through config-settings dotted keys anyway, so
  the feature is TOML-first by nature. The bare-string form must still work via
  env (`SKBUILD_FORCE_INCLUDE=src=dst;...`).

## Files to modify

### 1. `src/scikit_build_core/settings/skbuild_model.py`
Add a dataclass and a top-level field on `ScikitBuildSettings`:

```python
@dataclasses.dataclass
class ForceIncludeTargets:
    sdist: Optional[str] = None
    wheel: Optional[str] = None
    build: Optional[str] = None
    # each field gets a docstring (feeds configs.md + schema)

# on ScikitBuildSettings (top level, alongside `generate`, `metadata`):
force_include: Dict[str, Union[str, ForceIncludeTargets]] = dataclasses.field(
    default_factory=dict
)
```
Write a thorough field docstring covering the key=source rules, bare-string =
wheel-only, the inline table, the wheel leading-slash trees + experimental gate,
and the skip-if-missing behavior. `Dict`/`Union`/`Optional` already imported.

### 2. Settings machinery (the new general capability)
- **`sources.py` — `TOMLSource.convert()` union branch** (~line 726-738): if the
  item is a `dict` and one union member is a dataclass, dispatch to that
  dataclass's existing construction path (the `dataclasses.is_dataclass` branch
  at ~line 698). ~10-15 lines. Ensure a *string* item still resolves to the
  `str` member (bare-string form).
- **`sources.py` — `EnvSource.convert()` / `ConfSource.convert()`**: confirm a
  scalar value against `Union[str, dataclass]` resolves to `str` rather than
  hitting the "dataclasses not supported" guard (`get_inner_type` yields the
  Union, not the dataclass, so this likely already passes — add a test, tweak
  only if needed). Inline-table values are not expected from these sources.
- **`documentation.py`**: `get_display_type()` — handle non-Optional `Union`
  (render e.g. `str | ForceIncludeTargets`). `mk_docs()` — expand the dataclass
  member nested in `Dict[str, Union[str, dataclass]]` so `sdist`/`wheel`/`build`
  sub-fields are documented (required, else `nox -t gen` / `test_settings_docs`
  break). ~20 lines.
- **`json_schema.py`**: verify only — should already produce
  `{type: object, patternProperties: {".+": {oneOf: [string, <ForceIncludeTargets>]}}}`.

### 3. `src/scikit_build_core/build/_pathutil.py`
Helpers (reuse existing `scantree`; import `EXCLUDE_LINES` from
`._file_processor`):

- `resolve_wheel_tree(dest, *, wheel_dirs, targetlib, experimental) -> tuple[Path, str]`
  — factor out the leading-slash logic currently inlined in `wheel.py:302-314`
  (`..` check, experimental gate, `wheel_dirs` key validation); return the base
  tree dir and remaining relative dest. Reuse it for `wheel.install_dir` too.
- `iter_force_include(source, dest, base) -> Iterator[tuple[Path, Path]]` —
  yields `(src_file, base/dest[/relpath])` pairs. File source → one pair; dir
  source → `scantree` filtered by a `GitIgnoreSpec.from_lines(EXCLUDE_LINES)`;
  **missing source → yields nothing + `logger.debug`** (no raise). Shared by all
  three targets.

### 4. `src/scikit_build_core/build/wheel.py`
Normalize each `force_include` value to `(sdist, wheel, build)` dests (bare
string → wheel only).
- **wheel**: after `packages_to_file_mapping()` (line 458-466), for each entry
  with a wheel dest, `base = resolve_wheel_tree(wheel_dest, ...)`, then
  `iter_force_include(source, rest, base)`; copy each pair **always** (not gated
  on `if not editable`) and **after** the package-copy loop (468-471) so it
  overrides. Keep this out of the `mapping` returned/passed to `_make_editable`.
  `wheel.build()` (line 476) picks them up from `wheel_dirs`.
- **build**: inside `if cmake is not None:`, immediately **before**
  `builder.configure()` (line 411), copy each entry's build dest into `build_dir`
  via `iter_force_include(source, build_dest, build_dir)`.
- Reuse `resolve_wheel_tree` for the existing `install_dir` block (302-314).

### 5. `src/scikit_build_core/build/sdist.py`
After the `each_unignored_file` loop (151-156), for each `force_include` entry
with an sdist dest, `iter_force_include(source, sdist_dest, Path())` and
`tar.add(src_file, arcname=srcdirname / target, filter=normalize_tar_info if
reproducible else lambda x: x)`. Skip-missing handled in the helper.

### 6. Regenerate derived files
`nox -t gen` → `scikit-build.schema.json`, `docs/reference/configs.md`, README
cog. CI checks `git diff --exit-code`.

### 7. Narrative docs + changelog
Add a `force-include` section near the `wheel.packages` / `sdist.include` /
`install-dir` docs (search `docs/` for `wheel.exclude`), covering the three
targets, bare-string=wheel-only, leading-slash trees, override precedence, and
the from-sdist caveat. Add a changelog entry if `docs/changelog.md` exists.

## Caveats to document
- Building a wheel **from an sdist**: external (`../`) sources won't exist and
  are silently skipped (assumed already at the destination). Use the inline
  table's `sdist` key to vendor them into the tarball so the destination is
  populated. Same behavior as hatchling.
- Wheel leading-slash multi-tree targeting requires `experimental = true`.
- Inline-table values are TOML-only (env can set bare-string/wheel form).

## Tests
- **`tests/test_skbuild_settings.py`**: add `force_include == {}` default; add
  TOML cases for both the bare-string and inline-table forms (assert the bare
  string normalizes to wheel-only and the table populates `ForceIncludeTargets`);
  add an env/config-settings case for the bare-string form.
- **`tests/test_settings_overrides.py`**: confirm `force_include` dict merges
  across sources/overrides.
- **`tests/test_schema.py` / `tests/test_settings_docs.py`**: pass after
  regeneration (the `oneOf` schema + nested-dataclass docs).
- **New `tests/test_force_include.py`** (use existing wheel/sdist build
  fixtures + `tmp_path`): a single file, a recursed directory, an outside-tree
  (`../`) source, a leading-slash `/scripts` wheel dest, a `{build=...}` entry
  visible to CMake at configure, and a `{sdist=...}` entry landing in the
  tarball; assert `__pycache__`/`*.pyc` are skipped, force-include overrides a
  same-path package file, and a **missing source is skipped without error**.

## Verification
- `nox -t gen` then `git diff --exit-code` (schema/docs in sync).
- `prek -a --quiet` (ruff + mypy; use `scikit_build_core._compat` shims, no
  banned imports; `Dict[str, Union[str, ForceIncludeTargets]]` stays
  mypy-strict-clean).
- `nox -s tests -- -k force_include`.
- Manual: a throwaway package with a bare-string and an inline-table entry; run
  `uv run python -m build`, then `unzip -l` the wheel and `tar tzf` the sdist to
  confirm placement.


</p>
</details> 

<details><summary>:robot: Impl :robot:</summary>
<p>

`force-include` is implemented and all tests + linters pass. Here's what landed on the `feat/force-include` branch:

## What was built

A top-level `[tool.scikit-build.force-include]` table mapping source paths → destinations, modeled on hatchling but extended for scikit-build-core's multi-tree wheel layout.

```toml
[tool.scikit-build.force-include]
"vendor/lib.so"   = "pkg/_lib.so"                      # bare string → wheel (platlib)
"tools/run.sh"    = {wheel = "/scripts/run.sh"}        # leading-slash → wheel tree
"../shared/data"  = {sdist = "data", wheel = "pkg/data"}  # per-target, recursive dir
"toolchain.cmake" = {build = "toolchain.cmake"}        # into CMake build dir pre-configure
```

**Semantics:** bare string = wheel-only; inline table targets any subset of `sdist`/`wheel`/`build`. Sources may be files or directories (recursed, skipping VCS/`__pycache__` junk), can live outside the project (`../`, `~`), override package files/CMake output, and a **missing source is silently skipped** (your hatchling-compat correction — assumed already present when building from an sdist).

## Key implementation points

- **Settings machinery** — taught `TOMLSource.convert()` to route a TOML table to a dataclass member of a union (`Dict[str, Union[str, ForceIncludeTargets]]`), and `documentation.py` to render non-optional unions. `json_schema.py` already produced the right `oneOf`; only a one-line guard in `skbuild_schema.py` was needed for the `patternProperties` shape. Bare-string form works via env/config-settings too; the inline-table form is TOML-only.
- **`_pathutil.py`** — added `iter_force_include` (recursion + skip-missing) and factored the leading-slash wheel-tree resolution into `resolve_wheel_tree`, **reused by both `wheel.install-dir` and the hatch plugin** (removed a triplicated block).
- **Tests** — `tests/test_force_include.py` covers file/dir/external/leading-slash/override/missing/sdist/build-into-CMake, plus settings-parsing tests across all three sources.
- **Docs/schema** regenerated via `nox -t gen`; a narrative "Force-including files" section added.

I did **not** commit (you commit when ready) and **skipped the changelog** — it's release-generated with merged-PR references, so the PR itself is the right place for that entry.

One note for the PR: the inline-table form is TOML-only by design (force-include keys are file paths with dots/slashes that don't round-trip through config-settings dotted keys) — worth calling out so it's a deliberate, documented limitation rather than a surprise.

</p>
</details> 

<details><summary>:robot: GPT 5.5 review :robot:</summary>
<p>

The new force-include feature has correctness gaps: non-wheel destinations can escape their intended roots, and Hatch builds accept but ignore the new setting. These should be fixed or explicitly rejected before considering the patch correct.

Full review comments:

- [P2] Constrain non-wheel force-include destinations — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/_pathutil.py:136-142
  When a `{sdist = ...}` or `{build = ...}` destination is absolute or contains parent components, `base / dest` can escape the documented root. For example, `{build = "/tmp/toolchain.cmake"}` writes outside the CMake build directory, and an sdist destination like `../x` creates tar entries outside the sdist root. Since these destinations are documented as relative, validate/reject absolute paths and `..` before joining.

- [P2] Apply or reject force-include for Hatch builds — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/hatch/plugin.py:222-228
  The Hatch hook now accepts the new `settings.force_include` option, but this path never copies `{build = ...}` entries before configure or adds wheel targets to Hatch's `build_data`. A project using the new setting with the hatchling plugin will silently miss those files, unlike the PEP 517 backend. Please either mirror the backend `iter_force_include` handling here or reject the setting in `_validate` as unsupported.

</p>
</details>

PR #1249, opened 2026-03-03
feat: add "explicit" inclusion-mode (opt-in include)

### LINKED ISSUE #1248: Missing files in resulting package
When I clone ase2sprkkr package
```
git clone ase2sprkkr
cd ase2sprkkr
git checkout 047223aedc0e06ba678ae93c8368572a13ab38fe
```
and install it using
`pip install --no-deps .`
using scikit-build-core version 0.12.1, there are many missing files in the resulting installed package (in site_packages), e.g.
the whole subpackages 
```
ase2sprkkr.ase
ase2sprkkr.common.grammar_types
ase2sprkkr.bindings.nomad
```
or e.g. half of the examples from
`ase2sprkkr.examples`

If I pin the build backend to 0.11.5, the package is installed correctly.
There is no apparent reason, why some of the directories are excluded and some of them not.

henryiii:

Closes #1248.

Adds a fourth `sdist.inclusion-mode` value, **`"explicit"`**, for an allowlist-style ("opt-in") workflow.

### Behavior

Nothing is packaged unless it matches an `include` pattern, and `exclude` is applied **after** the opt-in, so it trims included files back out. This is the inverse of the existing modes, where `include` always wins over `exclude`. Like `"manual"`, git ignore files are not consulted.

| mode | reads gitignore | nested ignore | prunes dirs | default = include? | include vs exclude |
|------|------|------|------|------|------|
| classic  | yes | yes | no  | yes | include wins |
| default  | yes | yes | yes | yes | include wins |
| manual   | no  | no  | yes | yes | include wins |
| **explicit** | **no** | **no** | **yes** | **no (opt-in)** | **exclude wins** |

The mode is read from `[tool.scikit-build.sdist]` but governs both sdist file selection and wheel *package* collection (both paths pass `settings.sdist.inclusion_mode`), so in `"explicit"` mode a package directory contributes only files matched by `sdist.include`.

### Example

```toml
[tool.scikit-build.sdist]
inclusion-mode = "explicit"
include = ["src/**/*.py", "README.md"]
exclude = ["src/**/_scratch.py"]  # trims back out, even though include matched
```

### Implementation

- `settings/skbuild_model.py` — new `Literal` value + docstring (drives schema/docs generation).
- `build/_file_processor.py` — `each_unignored_file` skips git-ignore data for `"explicit"` (like `"manual"`) while still pruning directories; `match_path` gains an `explicit` branch with *exclude → include → exclude-by-default* precedence.
- `build/_pathutil.py` — widened the `mode` literal used for wheel package collection.
- Regenerated `scikit-build.schema.json` and `docs/reference/configs.md`.

### Tests

- `tests/test_file_processor.py` — dedicated tests for the opt-in baseline, exclude-wins-over-include (contrasted against `"manual"`), and git-ignore being skipped.
- `tests/test_skbuild_settings.py` — `SKBUILD_SDIST_INCLUSION_MODE=explicit` round-trip.

### DIFF
--- a/docs/reference/configs.md
+++ b/docs/reference/configs.md
@@ -456,7 +456,9 @@ print(mk_skbuild_docs())
 
   Files to include in the SDist even if they are skipped by default. Supports gitignore syntax.
 
-  Always takes precedence over :confval:`sdist.exclude`
+  In ``"default"`` and ``"classic"``, this takes precedence over
+  :confval:`sdist.exclude`. In ``"manual"``, files must be included first,
+  then exclude rules are applied.
 
   .. seealso::
      :confval:`sdist.exclude`
@@ -473,7 +475,8 @@ print(mk_skbuild_docs())
 
   * "default": Process the git ignore files. Shortcuts on ignored directories.
   * "classic": The behavior before 0.12, like "default" but does not shortcut directories.
-  * "manual": No extra logic, based on include/exclude only.
+  * "manual": Explicit allowlist mode; files must match ``include`` and then
+    are filtered by ``exclude``.
 
   If you don't set this, it will be "default" unless you set the minimum
   version below 0.12, in which case it will be "classic".
--- a/docs/reference/configs.md
+++ b/docs/reference/configs.md
@@ -681,7 +681,7 @@ print(mk_skbuild_docs())
 ```{eval-rst}
 .. confval:: sdist.inclusion-mode
 
-  :Type: ``"classic" | "default" | "manual"``
+  :Type: ``"classic" | "default" | "manual" | "explicit"``
   :Default: "default"  # "classic"
   :Config-settings: ``sdist.inclusion-mode`` or ``skbuild.sdist.inclusion-mode``
   :Environment variable: ``SKBUILD_SDIST_INCLUSION_MODE``
@@ -693,6 +693,9 @@ print(mk_skbuild_docs())
   * "default": Process the git ignore files. Shortcuts on ignored directories.
   * "classic": The behavior before 0.12, like "default" but does not shortcut directories.
   * "manual": No extra logic, based on include/exclude only.
+  * "explicit": Opt-in only. Nothing is included unless it matches an ``include``
+    pattern, and ``exclude`` is applied after, so it can trim included files back
+    out. Like "manual", git ignore files are not read. (1.0+)
 
   If you don't set this, it will be "default" unless you set the minimum
   version below 0.12, in which case it will be "classic".
--- a/src/scikit_build_core/settings/skbuild_model.py
+++ b/src/scikit_build_core/settings/skbuild_model.py
@@ -213,7 +213,9 @@ class SDistSettings:
     """
     Files to include in the SDist even if they are skipped by default. Supports gitignore syntax.
 
-    Always takes precedence over :confval:`sdist.exclude`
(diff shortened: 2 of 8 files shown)


docs/reference/configs.md
```diff
@@ -693,6 +693,9 @@ print(mk_skbuild_docs())
   * "default": Process the git ignore files. Shortcuts on ignored directories.
   * "classic": The behavior before 0.12, like "default" but does not shortcut directories.
   * "manual": No extra logic, based on include/exclude only.
+  * "explicit": Opt-in only. Nothing is included unless it matches an ``include``
+    pattern, and ``exclude`` is applied after, so it can trim included files back
+    out. Like "manual", git ignore files are not read. (0.13+)
```
henryiii:
```suggestion
    out. Like "manual", git ignore files are not read. (1.0+)
```

src/scikit_build_core/settings/skbuild_model.py
```diff
@@ -320,6 +320,9 @@ class SDistSettings:
     * "default": Process the git ignore files. Shortcuts on ignored directories.
     * "classic": The behavior before 0.12, like "default" but does not shortcut directories.
     * "manual": No extra logic, based on include/exclude only.
+    * "explicit": Opt-in only. Nothing is included unless it matches an ``include``
+      pattern, and ``exclude`` is applied after, so it can trim included files back
+      out. Like "manual", git ignore files are not read. (0.13+)
```
henryiii:
```suggestion
      out. Like "manual", git ignore files are not read. (1.0+)
```

henryiii:

The new mode mostly works, but it is not guarded by the documented minimum-version requirement. This can produce configurations that claim compatibility with older scikit-build-core releases that cannot parse the new value.

Review comment:

- [P2] Require minimum-version 0.13 for explicit mode — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/settings/skbuild_model.py:323-325
  When a project sets `sdist.inclusion-mode = "explicit"` while declaring `minimum-version = "0.12"`, this new value is accepted even though the docs mark it as 0.13+ and scikit-build-core 0.12 does not understand it. That lets projects advertise compatibility with 0.12 but then fail in build environments that legitimately install a 0.12 backend; the existing `sdist.inclusion-mode` minimum-version check should be tightened for this new enum value.

src/scikit_build_core/settings/skbuild_model.py
henryiii: :robot: _AI text below_ :robot:

Done — updated to `(1.0+)` in the model docstring and regenerated the docs.

docs/reference/configs.md
henryiii: :robot: _AI text below_ :robot:

Done — updated to `(1.0+)` in the model docstring and regenerated the docs.

henryiii:

Addressed in c0e671c. The `sdist.inclusion-mode` gate in `skbuild_read_settings.py` now additionally errors when `inclusion-mode = "explicit"` is combined with `minimum-version < 1.0`, so a project cannot advertise compatibility with a release that does not understand the new value. (Using `1.0` rather than `0.13`, since that is the next version.) Added `test_sdist_inclusion_mode_explicit_requires_minimum_version` to cover it.

Issue #1261, opened 2026-04-10
override_only config-settings rejected as hard-coded


yosuke-konno-0414:
Hi, I think there is a validation bug around `override_only` settings when they are provided via PEP 517 config-settings.

I expected `override_only` fields such as `cmake.toolchain-file` and `wheel.tags` to be accepted when passed via `-C...`, but they are rejected with:

```text
ERROR: cmake.toolchain-file is not allowed to be hard-coded in the pyproject.toml file
```

and similarly for `wheel.tags`.

## Expected behavior

Settings marked as `override_only=True` should be:

- rejected when hard-coded in `[tool.scikit-build]`
- accepted when provided via:
  - `[[tool.scikit-build.overrides]]`
  - PEP 517 config-settings such as `-Ccmake.toolchain-file=...`

## Actual behavior

When passing an `override_only` setting via config-settings, scikit-build-core appears to validate it as if it had been hard-coded in `pyproject.toml`.

## Minimal repro

`pyproject.toml`:

```toml
[build-system]
requires = ["scikit-build-core==0.12.2"]
build-backend = "scikit_build_core.build"

[project]
name = "repro"
version = "0.0.1"
```

Then run:

```bash
uv build --wheel -Ccmake.toolchain-file=foo.cmake
```

Observed error:

```text
ERROR: cmake.toolchain-file is not allowed to be hard-coded in the pyproject.toml file
```

I also see the same kind of failure with:

```bash
uv build --wheel -Cwheel.tags=cp312-abi3-win_amd64
```

## Why this looks like a bug

From reading the validation code, it looks like `_validate_overrides()` does:

```python
original_value = record.original_value if record else value
```

So if a setting came from config-settings, there is no override record, and the final merged `value` is treated as though it were the original pyproject value.

That seems to conflate:

- a value that was actually present in `pyproject.toml`
- a value that came from config-settings (`-C...`)

In other words, config-settings for `override_only` fields seem to be rejected because their source is not tracked separately from pyproject values during validation.

## Environment

- scikit-build-core: 0.12.2
- frontend: `uv build`

LecrisUT:
Hmm, good point. Afaik at that point the configs are already merged and we lost the source of those. Irc the sources are first merged when they are still raw dicts, but we are handling the overrides when they are already a dataclass. Ideas on how to approach this welcome

PR #1262, opened 2026-04-10
fix: detect override vs. config-settings

henryiii:
Fix for #1261 generated by copilot in VSCode using auto model selection.

Override-only validation checks the TOML-only snapshot instead of the fully merged settings, which stops config-settings from being misclassified as hard-coded pyproject values.

### DIFF
--- a/src/scikit_build_core/settings/skbuild_read_settings.py
+++ b/src/scikit_build_core/settings/skbuild_read_settings.py
@@ -137,13 +137,16 @@ def _handle_move(
 
 def _validate_overrides(
     settings: ScikitBuildSettings,
+    static_settings: ScikitBuildSettings,
     overrides: dict[str, OverrideRecord],
+    config_setting_keys: set[str],
 ) -> None:
     """Validate all fields with any override information."""
 
     def validate_field(
         field: dataclasses.Field[Any],
         value: Any,
+        static_value: Any,
         prefix: str = "",
         record: OverrideRecord | None = None,
     ) -> None:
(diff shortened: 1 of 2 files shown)


src/scikit_build_core/settings/skbuild_read_settings.py
```diff
@@ -137,22 +137,28 @@ def _handle_move(
...
         # Check if we had a hard-coded value in the record
         conf_key = field.name.replace("_", "-")
         if field.metadata.get("override_only", False):
-            original_value = record.original_value if record else value
-            if original_value is not None:
+            full_key = f"{prefix}{conf_key}"
+            original_value = record.original_value if record else static_value
+            if original_value is not None or (
+                value is not None and full_key not in config_setting_keys
+            ):
```
LecrisUT, line 161: I don't think the logic here is right, but can't tell exactly what it should be. I think it fails if the user provided if from env-variable.

src/scikit_build_core/settings/skbuild_read_settings.py
```diff
@@ -259,7 +274,7 @@ def __init__(
         )
         self.settings = self.sources.convert_target(ScikitBuildSettings)
 
-        static_settings = SourceChain(
+        self.static_settings = SourceChain(
```
LecrisUT, line 277: Maybe what we need here is to break down the `self.settings` from above to get the EnvSource/ConfSource separate from the `*toml_srcs` and in the other part check explicitly if the key is in the `dynamic_settings`.

henryiii:
Yes, this was just directly from a model. I think we either need to ensure env-vars are included, or refactor to separate handling a bit.

PR #1269, opened 2026-04-22
feat: support PEP 803, free-threaded tag

### LINKED ISSUE #1252: Explore support for abi3t ABI tag
[PEP 803](https://peps.python.org/pep-0803/) proposes a new `abi3t` ABI tag.

I've added preliminary support for `abi3t` in [packaging](https://github.com/pypa/packaging/pull/1099) and [setuptools](https://github.com/pypa/setuptools/pull/5193/changes#diff-5f0d2f159393c18a35b50b6a97212ace4cb44ef75df0c85dd28edb23dce9a782L356) and @mgorny added preliminary support in [meson-python](https://github.com/mesonbuild/meson-python/compare/main...mgorny:meson-python:freethreading-limited-api).

The `builder` package will need to be updated, presumably by adding new `abi3t` keyword arguments for `get_soabi` and `get_python_library` and by updating the wheel tag logic.

I think there would also need to be new abi3t example packages to go along side the abi3 packages.

We've been working on a test suite for packaging libraries and bindings generators, see https://github.com/Quansight-Labs/stable-abi-testing.

henryiii:
Support should come out in CMake 4.4, see https://gitlab.kitware.com/cmake/cmake/-/merge_requests/11955.

Fix #1252.

:robot: Assisted-by: Copilot:Kimi-K2.6

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -374,8 +374,8 @@ Scikit-build-core will only target ABI3 if the version of Python is equal to or
 newer than the one you set. `${SKBUILD_SABI_COMPONENT}` is set to
 `Development.SABIModule` when targeting ABI3 or ABI3T, and is an empty string
 otherwise. For free-threaded Python (PEP 703), you can use `cp315t` to target
-the free-threaded stable ABI, which sets `Py_TARGET_ABI3T` instead of
-`Py_LIMITED_API`.
+the free-threaded stable ABI, which sets `Py_TARGET_ABI3T` (if using CMake
+4.4+).
 
 If you are not using CPython at all, you can specify any version of Python is
 fine:
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -372,7 +372,10 @@ wheel.py-api = "cp38"
 
 Scikit-build-core will only target ABI3 if the version of Python is equal to or
 newer than the one you set. `${SKBUILD_SABI_COMPONENT}` is set to
-`Development.SABIModule` when targeting ABI3, and is an empty string otherwise.
+`Development.SABIModule` when targeting ABI3 or ABI3T, and is an empty string
+otherwise. For free-threaded Python (PEP 703), you can use `cp315t` to target
+the free-threaded stable ABI, which sets `Py_TARGET_ABI3T` instead of
+`Py_LIMITED_API`.
 
 If you are not using CPython at all, you can specify any version of Python is
 fine:
--- a/src/scikit_build_core/builder/builder.py
+++ b/src/scikit_build_core/builder/builder.py
@@ -210,7 +210,8 @@ class Builder:
             cache_config["SKBUILD_PROJECT_VERSION_FULL"] = str(version)
 
         py_api = self.settings.wheel.py_api
-        ft_abi = False
+        gil_disabled = bool(sysconfig.get_config_var("Py_GIL_DISABLED"))
+        ft_abi = limited_api is True and gil_disabled
         if limited_api is None:
             if py_api.startswith("cp3"):
                 if py_api.endswith("t"):
@@ -218,7 +219,7 @@ class Builder:
                     target_minor_version = int(py_api[3:-1])
                     if (
                         sys.implementation.name == "cpython"
-                        and sysconfig.get_config_var("Py_GIL_DISABLED")
+                        and gil_disabled
                         and target_minor_version <= sys.version_info.minor
                     ):
                         ft_abi = True
@@ -236,7 +237,7 @@ class Builder:
                     if sys.implementation.name != "cpython":
                         limited_api = False
                         logger.info("PyPy doesn't support the Limited API, ignoring")
-                    elif sysconfig.get_config_var("Py_GIL_DISABLED"):
+                    elif gil_disabled:
                         limited_api = False
                         logger.info(
                             "Free-threaded Python doesn't support the classic Limited API, ignoring"
@@ -252,7 +253,7 @@ class Builder:
             ft_abi = False
             logger.info("PyPy doesn't support the Limited API, ignoring")
 
-        if limited_api and sysconfig.get_config_var("Py_GIL_DISABLED") and not ft_abi:
+        if limited_api and gil_disabled and not ft_abi:
             limited_api = False
             logger.info(
                 "Free-threaded Python doesn't support the classic Limited API, ignoring"
--- a/src/scikit_build_core/builder/builder.py
+++ b/src/scikit_build_core/builder/builder.py
@@ -209,27 +209,41 @@ class Builder:
             )
             cache_config["SKBUILD_PROJECT_VERSION_FULL"] = str(version)
 
+        py_api = self.settings.wheel.py_api
+        ft_abi = False
         if limited_api is None:
-            if self.settings.wheel.py_api.startswith("cp3"):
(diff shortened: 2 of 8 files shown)


docs/configuration/index.md
```diff
@@ -372,7 +372,10 @@ wheel.py-api = "cp38"
 
 Scikit-build-core will only target ABI3 if the version of Python is equal to or
 newer than the one you set. `${SKBUILD_SABI_COMPONENT}` is set to
-`Development.SABIModule` when targeting ABI3, and is an empty string otherwise.
+`Development.SABIModule` when targeting ABI3 or ABI3T, and is an empty string
+otherwise. For free-threaded Python (PEP 703), you can use `cp315t` to target
+the free-threaded stable ABI, which sets `Py_TARGET_ABI3T` instead of
+`Py_LIMITED_API`.
```
henryiii:
```suggestion
`Py_LIMITED_API` (if using CMake 4.4+).
```

src/scikit_build_core/builder/builder.py
```diff
@@ -209,27 +209,41 @@ def configure(
...
         if limited_api is None:
-            if self.settings.wheel.py_api.startswith("cp3"):
-                target_minor_version = int(self.settings.wheel.py_api[3:])
+            if py_api.startswith("cp3") and py_api.endswith("t"):
+                target_minor_version = int(py_api[3:-1])
+                ft_abi = (
+                    sys.implementation.name == "cpython"
+                    and sysconfig.get_config_var("Py_GIL_DISABLED")
+                    and target_minor_version <= sys.version_info.minor
+                )
```
LecrisUT: Imo this would read clearer if it was an inner if check of `py_api.startswith("cp3")` part. Also could use some warning if a `cp3*t` is requested, but the other checks here fail

src/scikit_build_core/builder/builder.py
```diff
@@ -209,27 +209,41 @@ def configure(
...
+            ft_abi = False
             logger.info("PyPy doesn't support the Limited API, ignoring")
 
-        if limited_api and sysconfig.get_config_var("Py_GIL_DISABLED"):
+        if limited_api and sysconfig.get_config_var("Py_GIL_DISABLED") and not ft_abi:
             limited_api = False
             logger.info(
-                "Free-threaded Python doesn't support the Limited API currently, ignoring"
+                "Free-threaded Python doesn't support the classic Limited API, ignoring"
             )
```
LecrisUT: Can this be moved upwards? Seems like it is the warning that I referred to previously

src/scikit_build_core/builder/builder.py
```diff
@@ -209,27 +209,41 @@ def configure(
...
+            elif py_api.startswith("cp3"):
+                target_minor_version = int(py_api[3:])
                 limited_api = target_minor_version <= sys.version_info.minor
             else:
                 limited_api = False
 
         if limited_api and sys.implementation.name != "cpython":
             limited_api = False
+            ft_abi = False
             logger.info("PyPy doesn't support the Limited API, ignoring")
```
LecrisUT: This is a no-op afaict, but no objection to being more verbose, just that the info does not really reflect both options being set

src/scikit_build_core/builder/sysconfig.py
```diff
@@ -44,7 +44,9 @@ def __dir__() -> list[str]:
     return __all__
 
 
-def get_python_library(env: Mapping[str, str], *, abi3: bool = False) -> Path | None:
+def get_python_library(
+    env: Mapping[str, str], *, abi3: bool = False, abi3t: bool = False
+) -> Path | None:
```
LecrisUT, line 49: Instead of `abi3t` how about having a `ft` flag instead. It would make the previous usage simpler and less prone to dyslexia 🙂. Similarly for `get_soabi`

src/scikit_build_core/builder/builder.py
LecrisUT, line 154: What's the future of this variable? Any reason for `ft_abi` to also not have a similar logic?

src/scikit_build_core/builder/builder.py
```diff
@@ -265,20 +279,26 @@ def configure(
...
-        py_api = self.settings.wheel.py_api
-        cache_config["SKBUILD_SABI_VERSION"] = (
-            f"{py_api[2]}.{py_api[3:]}"
-            if limited_api and py_api.startswith("cp")
-            else ""
-        )
+        if limited_api and py_api.startswith("cp"):
+            version_str = py_api[2:]
+            if version_str.endswith("t"):
+                version_str = version_str[:-1]
```
LecrisUT, line 311: How about `remove_suffix`/`remove_prefix` instead?

src/scikit_build_core/builder/wheel_tag.py
```diff
@@ -99,30 +99,69 @@ def compute_best(
...
+                )
+                or (
+                    x.startswith("cp3")
+                    and len(x) > 4
+                    and x[3:-1].isdecimal()
+                    and x.endswith("t")
+                    and sysconfig.get_config_var("Py_GIL_DISABLED")
+                )
+                for x in pyvers_new
+            ):
```
LecrisUT: Oh my, this has grown quite unreadable 🙃. How about making a helper class out of it to interrogate these.

tests/test_builder.py
```diff
@@ -269,6 +275,68 @@ def test_wheel_tag_with_abi_darwin(monkeypatch):
...
+            self.releaselevel = releaselevel
+            self.serial = serial
+
+        def __getitem__(self, index: int) -> int | str:
+            return (self.major, self.minor, self.micro, self.releaselevel, self.serial)[
+                index
+            ]
+
+        def __ge__(self, other: tuple[int, ...]) -> bool:
+            return (self.major, self.minor, self.micro) >= other[:3]
```
LecrisUT: Oh look a helper class for the previous point 😅, just needs to take in a string to build itself.

src/scikit_build_core/builder/builder.py
henryiii: Python 3.9+ required.

src/scikit_build_core/builder/sysconfig.py
henryiii: This is officially called abi3t, so I think it's best to follow that. And ft sounds like free-threaded in general, not just the free-threaded stable ABI.

src/scikit_build_core/builder/builder.py
henryiii: I included ft_abi in the check too in the latest commit

src/scikit_build_core/builder/sysconfig.py
LecrisUT: Then how about allowing `abi3` to be arbitrary and parse `abi3t` logic first?

src/scikit_build_core/builder/builder.py
henryiii: I think this one variable is enough - it should enforce Limited API on either version of Python. But currently, it doesn't work for both.

The reason for it is py_limited_api from setuptools, which looks like it's just getting generalized in https://github.com/pypa/setuptools/pull/5193, so I think we should just fix the logic here.

henryiii: Fixed, this should now enable free-threaded mode on either. Might not work correctly until the setuptools PR goes in, though.

ngoldbaum:
Thanks for working on this!

PR #1272, opened 2026-04-23
bug: Properly handle Union in file-api

LecrisUT:
Previously only the first type was considered, but evidently that can potentially fail. Seems like this would only happen in one case

https://github.com/scikit-build/scikit-build-core/blob/db72d462f75f25aa372ab2c856106f322976a870/src/scikit_build_core/file_api/model/directory.py#L25

### DIFF
--- a/src/scikit_build_core/file_api/model/index.py
+++ b/src/scikit_build_core/file_api/model/index.py
@@ -57,10 +57,10 @@ class CMake:
 
 @dataclasses.dataclass(frozen=True)
 class Reply:
-    codemodel_v2: Optional[CodeModel]
-    cache_v2: Optional[Cache]
-    cmakefiles_v1: Optional[CMakeFiles]
-    toolchains_v1: Optional[Toolchains]
+    codemodel_v2: Optional[CodeModel] = None
+    cache_v2: Optional[Cache] = None
+    cmakefiles_v1: Optional[CMakeFiles] = None
+    toolchains_v1: Optional[Toolchains] = None
 
 
 @dataclasses.dataclass(frozen=True)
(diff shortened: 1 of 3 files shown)


src/scikit_build_core/file_api/model/index.py
```diff
@@ -57,10 +57,10 @@ class CMake:
...
 @dataclasses.dataclass(frozen=True)
 class Reply:
-    codemodel_v2: Optional[CodeModel]
-    cache_v2: Optional[Cache]
-    cmakefiles_v1: Optional[CMakeFiles]
-    toolchains_v1: Optional[Toolchains]
+    codemodel_v2: Optional[CodeModel] = None
+    cache_v2: Optional[Cache] = None
+    cmakefiles_v1: Optional[CMakeFiles] = None
+    toolchains_v1: Optional[Toolchains] = None
```
LecrisUT, line 63: Because this part is also touching the logic of `Optional` I think we should check what exactly should be done here?

These very much depend on the query used. Any suggestion on how these should be handled ideally?

One thing I would want is to drop the `_v` suffix there and make it completely version independent, but technically I think they are allowed to have multiple such fields. I think we could fix a specific `query` format that is allowed and that we parse

henryiii: Aren't the `_v1` actually part of the layout?

LecrisUT: Yes, but we can still use the `rename` for that. It is mostly cosmetic for us.

The bigger question is if we should assume there is only one variant. We should use the stateful queries and then we have our own namespace and exact control of the format we request and lock the `query`-`reply` format

henryiii (approved):

PR #1273, opened 2026-04-23
refactor: Move the file-api models to multi-versioned interface

LecrisUT:
- [ ] Convert the object kind dataclasses to versioned types
  - [ ] `codemodel`
  - [x] `configureLog`
  - [x] `cache`
  - [x] `cmakeFiles`
  - [x] `toolchains`
- [ ] Drop the logic for handling the previous approach
- [ ] Limit the exports to the main object/versioned aliases

Depends-on: #1012


src/scikit_build_core/file_api/model/toolchains.py
```diff
@@ -29,14 +35,53 @@ class Compiler:
...
+    commandFragment: Optional[str] = None
+
+
+@dataclasses.dataclass(frozen=True)
+class Toolchain_1_1(Toolchain_1_0):
+    """
+    ..versionadded:: 4.3.0
+    """
+
+    compiler: Compiler_1_1
```
LecrisUT, line 74: Looking at the format here with fresh eyes, this looks much less navigable then grouping them together. Afaik the fields are all additive in the same major although that is not stated as being a guarantee.

The main good thing here is the documentation of the fields and what version of the object kind and cmake version they are introduced.

@henryiii what are your thoughts? I am considering scrapping this approach and do only the documentation, making sure the object kind/version is easily introspected, and generalizing the loading of indirect objects (from `jsonFile` or equivalents)

PR #1276, opened 2026-04-29
docs: improve docs on source

henryiii:
I used copilot to fill out some docs for this file, as it's always been tricky to come back in and work through how it works.

### DIFF
--- a/src/scikit_build_core/settings/sources.py
+++ b/src/scikit_build_core/settings/sources.py
@@ -1,12 +1,12 @@
 """
-This is the configuration tooling for scikit-build-core. This is built around
-the :class:`Source` Protocol. Sources are created with some input (like a toml
-file for the :class:`TOMLSource`). Sources also usually have some prefix (like
-``tool.scikit-build``) as well. The :class:`SourceChain` holds a collection of
-Sources, and is the primary way to use them.
+This module implements the configuration sources used by scikit-build-core.
+Each concrete :class:`Source` adapts one input representation, such as
+environment variables, PEP 517 ``config-settings``, or nested TOML tables, into
+the same nested dataclass model. :class:`SourceChain` combines those sources and
+applies them in precedence order when building the final settings object.
 
-An end user interacts with :class:`SourceChain` via ``.convert_target``, which
-takes a Dataclass class and returns an instance with fields populated.
+An end user usually interacts with :class:`SourceChain`, which takes a
+dataclass type and returns an instance with fields populated.
 
 Example of usage::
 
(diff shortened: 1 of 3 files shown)


src/scikit_build_core/settings/sources.py
```diff
@@ -172,7 +217,14 @@ def all_option_names(self, target: type[Any]) -> Iterator[str]:
 
 class EnvSource:
     """
-    This is a source using environment variables.
+    Source backed by environment variables.
+
+    Native representation:
+
+    - ``has_item("a", "b")`` looks for ``PREFIX_A_B``.
+    - ``get_item`` always returns the raw environment variable string.
```
LecrisUT: Be careful of what the generated code is saying

src/scikit_build_core/settings/sources.py
```diff
@@ -27,6 +27,21 @@
...
+- :class:`ConfSource`: ``settings`` is a flat mapping keyed by dotted option
+  names. ``get_item`` returns a string/list/bool for scalar options, or builds
+  a ``dict[str, str]`` from ``option.key=value`` entries when the target field
+  is a dict.
+- :class:`TOMLSource`: ``settings`` is a nested mapping. ``get_item`` returns
+  the native TOML value at that path, such as a string, bool, list, dict, or a
+  nested table.
+- :class:`SourceChain`: reads a source-native value via ``get_item`` and then
+  immediately hands it back to that same source's ``convert`` method to produce
+  the final dataclass field value.
```
LecrisUT: The reference to the functions seems odd in this one. Shouldn't this document what the classes represent and how they connect to each other?

src/scikit_build_core/settings/sources.py
```diff
@@ -281,11 +340,17 @@ def _unrecognized_dict(
...
-    nested dicts. "verify" is a boolean that determines whether unrecognized
-    options should be checked for. Only set this to false if this might be sharing
-    config options at the same level.
+    Source for PEP 517 ``config-settings``.
+
+    ``settings`` is the flat mapping passed to the backend, where keys are
+    dotted option names like ``"cmake.args"``. Scalar values are stored
+    directly. Dict-typed target fields are represented by multiple flat entries
+    such as ``"sdist.include.foo" = "bar"``, and ``get_item(..., is_dict=True)``
+    reconstructs those suffixes into ``{"foo": "bar"}``.
```
LecrisUT: Should be moved to attribute documentation instead?

src/scikit_build_core/settings/sources.py
```diff
@@ -172,7 +217,14 @@ def all_option_names(self, target: type[Any]) -> Iterator[str]:
...
     """
-    This is a source using environment variables.
+    Source backed by environment variables.
+
+    Native representation:
+
+    - ``has_item("a", "b")`` looks for ``PREFIX_A_B``.
+    - ``get_item`` always returns the raw environment variable string.
+    - ``convert`` is responsible for parsing that string into the requested
+      field type.
```
LecrisUT: Isn't this duplicating the documentation of the protocol? Why don't we subclass the protocol instead and inherit the docstring?

src/scikit_build_core/settings/sources.py
```diff
@@ -76,6 +91,12 @@ def __dir__() -> list[str]:
 
 
 def _dig_strict(_dict: Mapping[str, Any], /, *names: str) -> Any:
+    """
+    Walk a nested mapping and return the value at ``names``.
+
+    Each input ``name`` is one nesting level. Missing keys propagate the
+    underlying ``KeyError``.
```
LecrisUT:
```suggestion
    :raises KeyError: when ``names`` is missing
```
Could be less technical about the explicit implementation

src/scikit_build_core/settings/sources.py
```diff
@@ -86,12 +107,22 @@ def _process_bool(value: str) -> bool:
 
 
 def _dig_not_strict(_dict: Mapping[str, Any], /, *names: str) -> Any:
+    """
+    Walk a nested mapping like :func:`_dig_strict`, but stop missing branches
+    by returning ``{}`` for the rest of the walk.
```
LecrisUT: Similarly something like ", but return an empty dict `{}` if the `names` is missing"

PR #1277, opened 2026-04-29
chore: adapt to better pathspec types

henryiii:
After https://github.com/cpburnz/python-pathspec/issues/108 this now works better, so adapting to that.


LecrisUT (approved):
Cool

PR #1278, opened 2026-04-29
feat: support inplace editable for setuptools plugin

henryiii:
Adding support for inplace editable installs with the setuptools plugin.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -71,7 +71,7 @@ repos:
         files: ^(src|tests|noxfile.py)
         args: []
         additional_dependencies:
-          - build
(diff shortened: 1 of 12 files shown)


LecrisUT:
Is there at least an upstream issue about being able to provide our own `.pth` file that could be referenced in the docs?

henryiii:
I was focusing on supporting `inplace` in this PR. I'm not sure about installing a custom .pth. I'd like to implement that for hatchling (which does support it, I think), then look into if it also could work here.

henryiii:
Oh, and we should probably see what we might need to do if the pth replacement PEP gets accepted.

PR #1279, opened 2026-04-29
Fix importlib.resources.files() in redirect-mode editable installs (#807)

### LINKED ISSUE #807: Editable installs do not support traversing files in the build

henryiii:
Fix #807. Looking at what copilot does for this.

Assisted-by: Copilot:claude-sonnet-4.6

### DIFF
--- a/src/scikit_build_core/resources/_editable_redirect.py
+++ b/src/scikit_build_core/resources/_editable_redirect.py
@@ -83,7 +83,7 @@ class _SkbuildMultiplexedPath:
 
     Used to combine source-tree and CMake-install-tree paths so that
     importlib.resources.files() can see files from both locations.
-    Works on Python 3.8+ without requiring importlib.resources.readers.
+    Fallback for Python < 3.11, which lacks importlib.resources.readers.MultiplexedPath.
     """
 
     def __init__(self, *paths: str) -> None:
@@ -168,6 +168,10 @@ class _ScikitBuildEditableReader:
             return Path(self._paths[0]) if self._paths else Path(".")
         if len(existing) == 1:
             return Path(existing[0])
+        if sys.version_info >= (3, 11):
+            from importlib.resources.readers import MultiplexedPath
+
+            return MultiplexedPath(*existing)
         return _SkbuildMultiplexedPath(*existing)
 
 
--- a/src/scikit_build_core/resources/_editable_redirect.py
+++ b/src/scikit_build_core/resources/_editable_redirect.py
@@ -208,6 +208,36 @@ class _ScikitBuildResourceLoaderWrapper:
         return _ScikitBuildEditableReader(self._skbuild_paths)
 
 
+def _patch_importlib_resources_for_python39() -> None:
+    """
+    Make importlib.resources.files() honor the editable resource reader on Python 3.9.
+
+    Python 3.9 ignores get_resource_reader().files() and falls back to
+    Path(spec.origin).parent, which points at the source tree for editable
+    installs. Patch that fallback so redirect packages can expose build-tree
+    resources there too.
+    """
+
+    if sys.version_info[:2] != (3, 9):
+        return
+
+    from importlib.resources import _common  # type: ignore[attr-defined]
+
+    if getattr(_common, "_skbuild_editable_patched", False):
+        return
+
+    original_fallback_resources = _common.fallback_resources
+
+    def fallback_resources(spec: object) -> object:
+        loader = getattr(spec, "loader", None)
+        if isinstance(loader, _ScikitBuildResourceLoaderWrapper):
+            return loader.get_resource_reader(getattr(spec, "name", "")).files()
+        return original_fallback_resources(spec)
+
+    _common.fallback_resources = fallback_resources
+    _common._skbuild_editable_patched = True
+
+
 class ScikitBuildRedirectingFinder(importlib.abc.MetaPathFinder):
     def __init__(
         self,
@@ -412,6 +442,7 @@ def install(
     :param install_dir: The wheel install directory override, if one was
                         specified
     """
+    _patch_importlib_resources_for_python39()
     sys.meta_path.insert(
         0,
         ScikitBuildRedirectingFinder(
--- a/src/scikit_build_core/resources/_editable_redirect.py
+++ b/src/scikit_build_core/resources/_editable_redirect.py
@@ -235,7 +235,7 @@ def _patch_importlib_resources_for_python39() -> None:
         return original_fallback_resources(spec)
 
     _common.fallback_resources = fallback_resources
-    _common._skbuild_editable_patched = True
+    _common._skbuild_editable_patched = True  # pylint: disable=protected-access
 
 
 class ScikitBuildRedirectingFinder(importlib.abc.MetaPathFinder):
@@ -263,7 +263,7 @@ class ScikitBuildRedirectingFinder(importlib.abc.MetaPathFinder):
 
         # Construct the __path__ of all resource files
         # I.e. the paths of all package-like objects
-        submodule_search_locations: dict[str, list[str]] = {}
+        submodule_search_locations: dict[str, set[str]] = {}
         pkgs: list[str] = []
         # Loop over both python native source files and cmake installed ones
         for tree in (known_source_files, known_wheel_files):
@@ -278,7 +278,7 @@ class ScikitBuildRedirectingFinder(importlib.abc.MetaPathFinder):
                 if not parent:
                     continue
                 # Initialize the tree element if needed
-                submodule_search_locations.setdefault(parent, [])
+                submodule_search_locations.setdefault(parent, set())
                 # Add the parent path to the dictionary values
                 parent_path = os.path.dirname(file)
                 if not parent_path:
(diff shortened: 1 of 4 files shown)


src/scikit_build_core/resources/_editable_redirect.py
```diff
@@ -77,6 +77,120 @@ def release(self) -> None:
         os.close(self.lock_file_fd)
 
 
+class _SkbuildMultiplexedPath:
```
LecrisUT, line 80: From #807 and some other research, it is perfectly fine to use `MultiplexedPath` from either `importlib.resources` and `importlib_resources`, and similarly for all the other classes. It is fine if you inter-mix them.

Please let's use those as bases

henryiii: This will complicate it for 3.8, I think, so let's go ahead and drop 3.8 now?

(manylinux/cibuildwheel drops 3.8 on Monday)

LecrisUT: But for 3.8 we can add it from `importlib_metadata` can't we? Iirc we also need it for anything below 3.11 because of a bug with multiplexedpath, but can't remember the details.

henryiii: Yeah, I was not sure about 3.9-3.10. If we need it for 3.9-3.10, then it doesn't really change anything.

henryiii: One problem is that this is a user dependency, rather than scikit-build's dependency (as scikit-build isn't installed into the editable install). If we can't use `importlib.resources` for 3.9-3.10, though, then that doesn't make it simpler.

henryiii: (Testing a generated fix that doesn't use multiplexedpath, I'll rework it later but wanted to see if it passes)

LecrisUT: > Yeah, I was not sure about 3.9-3.10. If we need it for 3.9-3.10, then it doesn't really change anything.

I misremembered it was 3.11 inclusive, so even worse. [importlib-resources > 5.11](https://github.com/python/importlib_resources/blob/main/NEWS.rst#v5110) iirc

> One problem is that this is a user dependency, rather than scikit-build's dependency (as scikit-build isn't installed into the editable install). If we can't use `importlib.resources` for 3.9-3.10, though, then that doesn't make it simpler.

Yeah I mulled that over night. Technically we could inject it in the runtime dependency but that's intrusive. We could error out on import if it's not available and ask them to manually install, but this may be used in CI as well (code coverage comes to mind) and that would be annoying. Last option is to try import either of them and fallback to our own implementation. At least it would be easy to patch out afterwards.

`MultiplexedPath` itself I am not very worried about since it is not hard to replicate that one. But one of the `Reader` or `Loader` (don't recall which one) I am more concerned because that one has some logic of getting the source and should have different logic when loading a python file vs a compiled one (and even `__pycache__`, but I haven't dug into that part) which is much trickier

Edit: It is the [`SourceLoader`](https://docs.python.org/3/library/importlib.html#importlib.abc.SourceLoader) and equivalents that I am thinking of

src/scikit_build_core/resources/_editable_redirect.py
```diff
@@ -77,6 +77,167 @@ def release(self) -> None:
...
+
+
+def _patch_importlib_resources_for_python39() -> None:
+    """
+    Make importlib.resources.files() honor the editable resource reader on Python 3.9.
+
+    Python 3.9 ignores get_resource_reader().files() and falls back to
+    Path(spec.origin).parent, which points at the source tree for editable
+    installs. Patch that fallback so redirect packages can expose build-tree
+    resources there too.
```
LecrisUT, line 228: From what I can see the `fallback_resources` is called only in `from_package` which should still honor the `loader().files()`

https://github.com/python/importlib_resources/blob/90aa959c9bc740a3eb9cdc1c51d87ed43d8a8ed8/importlib_resources/_common.py#L19

Got some reading for this part?

src/scikit_build_core/resources/_editable_redirect.py
```diff
@@ -77,6 +77,167 @@ def release(self) -> None:
...
+
+    def is_resource(self, name: str) -> bool:
+        path = self.files().joinpath(name)  # type: ignore[attr-defined]
+        return hasattr(path, "is_file") and path.is_file()
+
+    def contents(self) -> object:
+        return (item.name for item in self.files().iterdir())  # type: ignore[attr-defined]
+
+
+class _ScikitBuildResourceLoaderWrapper:
```
LecrisUT, line 201: I believe there is some codepaths that check either for the type of the `Reader` or the `Loader`. It would be better to either subclass some abc of it. I was thinking `__instancecheck__` overwriting would be useful, but I believe that is doing the opposite of what we would want

src/scikit_build_core/resources/_editable_redirect.py
```diff
@@ -77,6 +77,167 @@ def release(self) -> None:
...
+        return f"_SkbuildMultiplexedPath({paths})"
+
+
+class _ScikitBuildEditableReader:
+    """
+    Resource reader for editable installs with multiple package roots.
+
+    Provides importlib.resources.files() access across both the source tree
+    and the CMake install tree.
+    """
```
LecrisUT, line 159: Looking at the implementation of this I am more baffled that our original approach with the `submodule_search_locations` does not work out-of-the-box :/

src/scikit_build_core/resources/_editable_redirect.py
```diff
@@ -126,7 +287,26 @@ def __init__(
...
                 if not os.path.isabs(parent_path):
                     parent_path = os.path.join(self.dir, parent_path)
-                submodule_search_locations[parent].add(parent_path)
+                if parent_path not in submodule_search_locations[parent]:
+                    submodule_search_locations[parent].append(parent_path)
+        # Second pass: propagate build-tree paths from parent packages to
+        # sub-packages.  This covers the case where a Python package (with
+        # __init__.py) lives in a directory that also contains CMake-generated
+        # data files but has no Python modules of its own in the build tree.
+        # Processing in depth order ensures parents are resolved before children.
```
LecrisUT: Could we split this into its own PR so we can narrow down its effect?

LecrisUT (commented):
Would take some time to cross-reference with importlib stuff because I want to understand what was the magic sauce that we are missing. But CI passes so 🎉 for it

src/scikit_build_core/resources/_editable_redirect.py
henryiii: We really shouldn't be adding dependencies if the user doesn't have `dynamic=["dependencies"]`, but for editable installs only we could probably make an exception if we had to. We could error though if it's not present, with a message, on editable installs.

LecrisUT: Vote for error with message for me. Hmm, but I can see that affecting some workflows, so might want to have a quick response to change it to a different model

src/scikit_build_core/resources/_editable_redirect.py
henryiii: Fed this to Claude, it split the PRs.

<details><summary>:robot: Claude Opus 4.8</summary>
<p>

Branch 1: henryiii/feat/import — committed & pushed (e584c2e)

Cleanup addressing the PR review:
- Deduplicated find_spec into a shared _make_spec() helper.
- stdlib MultiplexedPath on 3.12+ (lazy import, no startup cost), hand-rolled fallback for 3.9–3.11. Verified the boundary empirically — stdlib's MultiplexedPath mishandles single slash-arg joinpath("a/b") on 3.11 (LecrisUT's "3.11 inclusive" was exactly right), fixed in 3.12.
- Removed the second-pass propagation — it had no test coverage and broke nothing when removed.

Branch 2: henryiii/feat/import-subpkg-propagation — committed (f1b9143)

Reworked the second-pass propagation as its own focused change. The important discovery while building a demonstrating test: most data files don't need it, because is_valid_module() only checks the pre-dot segment is an identifier — so generated_data, data.json, even pkg.so all get registered as fake "modules," and the first pass already records their directory. The second pass is only needed when a data file's name is not a valid identifier (e.g. contains a hyphen).

So the new test installs nested2-generated-data.txt into the regular nested2 subpackage and asserts files("cmake_generated.nested2").joinpath(...) finds it. I verified via manual editable installs that the test fails without the second pass and passes with it on 3.11, 3.12, and 3.14 — it genuinely exercises the propagation, which the original code never did.

Worth noting in the PR thread: this narrow trigger condition (non-identifier data filenames) is the honest scope of what the second pass fixes — you may want to mention that when discussing it with LecrisUT.

</p>
</details>

vyasr:
> Would take some time to cross-reference with importlib stuff because I want to understand what was the magic sauce that we are missing. But CI passes so 🎉 for it

@LecrisUT if I'm reading this PR right, the magic sauce was just having a consistent way to describe the multiplexedpath, which @henryiii added here. #808 specifically called out [needing a Traversable requiring multiple paths](https://github.com/scikit-build/scikit-build-core/pull/808/changes#diff-18cd92ae657a64dd0965cd71cea8b308d10154d73d69e0e4891494c36431fb91R90), but I didn't think methods like `joinpath` would have a sensible implementation that wouldn't surprise users if a file could exist in multiple places.

PR #1280, opened 2026-04-29
chore: skip git worktree & better architecture

henryiii:
I'm using `git worktree` and that makes a `.git` file that we need to avoid putting into the SDist.

Also updating the AGENTS.md with more architecture info.

:robot: Assisted-by: Pi:Kimi-K2.6

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -175,3 +175,6 @@ Thumbs.db
 .vscode/
 uv.lock
 *pylock
+
+# Git worktrees
+.git
--- a/.gitignore
+++ b/.gitignore
@@ -175,6 +175,3 @@ Thumbs.db
 .vscode/
 uv.lock
 *pylock
-
-# Git worktrees
-.git
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -17,6 +17,7 @@ It is built with **hatchling** and stored under `src/scikit_build_core/`.
   - `nox -t gen` — run all code generators (cog for README + schema + config
     reference + API docs).
   - `nox -s minimums` — lowest-direct dependency tests.
+  - `nox -s pylint` - some slightly-slower lints not included in `prek`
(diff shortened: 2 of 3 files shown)


.gitignore
```diff
@@ -175,3 +175,6 @@ Thumbs.db
 .vscode/
 uv.lock
 *pylock
+
+# Git worktrees
+.git
```
LecrisUT: Oh that's interesting. I've only used subtrees not worktree. It reads quite weird to git ignore the `.git`, do you have some references on this? From what I read the git ignore they used was on `.worktrees`, but I guess you have it used on the path under vendor? Can it ignore a more specific path like those under vendor?

henryiii: If you make a worktree:

```
git branch -c henryiii/feat/import
git worktree add ../scikit-build-core-import henryiii/feat/import
cd ./scikit-build-core-import
```

Then the directory has a `.git` _file_ that contains the text: `gitdir: /Users/henryfs/git/scikit-build/scikit-build-core/.git/worktrees/scikit-build-core-import` (or wherever the actual `.git` resides for you).

I think hatchling should ignore this by default; we could also add it to the explicit ignores instead in the pyproject.toml - on reflection, that's probably better.

LecrisUT: Ah it's not about hinting for git to ignore that file but for hatchlings. Makes sense then, but yeah, would be less confusing in pyproject.toml

LecrisUT (approved):

.gitignore
henryiii: I usually try to use `.gitignore` vs. telling hatchling what to ignore, but this is a case where I think telling just hatchling is correct. I think hatchling should probably ignore this by default, I mentioned it in the PyPA discord.

LecrisUT: Also the contents of `.git/info/exclude`, but who's counting 😁. I believe I raised that issue already.

PR #1282, opened 2026-04-30
feat: hatchling edit

henryiii:
Editable support for the hatchling plugin.

### DIFF
--- a/src/scikit_build_core/build/_editable.py
+++ b/src/scikit_build_core/build/_editable.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import os
 import typing
+from collections.abc import Mapping
 from pathlib import Path
 
 from ..resources import resources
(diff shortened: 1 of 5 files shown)


henryiii:
@LecrisUT Any thoughts as to why just the one fedora test might be failing here? It's on the new code, so it looks valid.

LecrisUT:
Oh interesting. The only difference that should be there are the compilers. I'll need to dig into the code changes.

Btw `tmt` is meant to be replicable locally, but we do not have mac users or CI to test it on, but if you have docker/podman and tmt installed (even from pypi), the following would replicate that job
```console
$ tmt run -a \
  provision --how=container --image=fedora:rawhide \
  plans --name=/tests/plans/pytest \
  login -t
```

LecrisUT:
Ok, I am reproducing locally. My suspicion is that `EXT_SUFFIX` gets calculated a bit differently so the file is not as expected, but let me see with the debugger.

LecrisUT:
Oops, wrong finding previously:
- `ext_location = '/tmp/pytest-of-cle/pytest-2/test_hatchling_editable_instal0/venv/lib/python3.14/site-packages/hatchling_example/_core.cpython-314-x86_64-linux-gnu.so'`
- `expected = '/tmp/pytest-of-cle/pytest-2/test_hatchling_editable_instal0/venv/lib64/python3.14/site-packages/hatchling_example/_core.cpython-314-x86_64-linux-gnu.so'`

see the `lib` vs `lib64` difference there

Issue #319, opened 2023-05-18
feat(setuptools): cmake_with_sdist


henryiii:
Usage stats (2 / 374):

| Location | entry |
|----------|-------|
| [MerginMaps/geodiff:setup.py](https://github.com/MerginMaps/geodiff/blob/HEAD/setup.py) | `False` |
| [Socrats/EGTTools:setup.py](https://github.com/Socrats/EGTTools/blob/HEAD/setup.py) | `True` |

henryiii:
:robot: Copilot:Kimi-K2.6 plan

# Plan: #319 — Support `cmake_with_sdist` in setuptools plugin

## Problem

The setuptools compatibility wrapper (`scikit_build_core.setuptools.wrapper.setup`) currently asserts that `cmake_with_sdist` is not supported. This is a keyword argument from scikit-build (classic) that users migrating to scikit-build-core expect to work.

Reference: [scikit-build-core#319](https://github.com/scikit-build/scikit-build-core/issues/319)

## Proposed approach

When `cmake_with_sdist=True` is passed via `setup()`, we will wire it through to the native scikit-build-core `sdist.cmake` setting. This means:

1. The setuptools `setup()` keyword argument is accepted (remove the `assert`).
2. The value is propagated to the `ScikitBuildSettings` in the `SettingsReader`.
3. This affects **both** `get_requires_for_build_sdist` (injects cmake/ninja requirements) and the actual `build_sdist` logic (runs CMake before packaging), matching the native backend behavior exactly.

## Key files to change

1. **`src/scikit_build_core/setuptools/wrapper.py`**
   - Remove the `assert not cmake_with_sdist` line.
   - Pass `cmake_with_sdist=cmake_with_sdist` into the `setuptools.setup()` call.

2. **`src/scikit_build_core/setuptools/build_cmake.py`**
   - Add a `cmake_with_sdist` entry-point handler (like `cmake_source_dir`, `cmake_args`).
   - When `True`, set an attribute on the `Distribution` so the build-meta layer can read it.

3. **`src/scikit_build_core/setuptools/build_meta.py`**
   - In `get_requires_for_build_sdist`, read the distribution / detect `cmake_with_sdist` and include cmake/ninja requirements when true.
   - Alternatively, hook into `SettingsReader` so that `sdist.cmake` reflects the `cmake_with_sdist` value.

4. **`src/scikit_build_core/settings/skbuild_read_settings.py` or `build_cmake.py`**
[...]

PR #1287, opened 2026-04-30
feat(setuptools): Allow cmake_with_sdist=False

henryiii:
Fixing #319.

This is only set to False now, so let's just only support False or missing.

:robot: Assisted-by: Copilot:Qwen-3.5:397B


src/scikit_build_core/setuptools/build_cmake.py
```diff
@@ -414,3 +414,14 @@ def cmake_install_target(
...
+
+def cmake_with_sdist(
+    _dist: Distribution,
+    attr: Literal["cmake_with_sdist"],
+    value: bool,  # noqa: FBT001
+) -> None:
+    assert attr == "cmake_with_sdist"
+    if value:
+        msg = "cmake_with_sdist must not be set to True"
+        raise setuptools.errors.SetupError(msg)
```
LecrisUT, line 427: Even if it is set to False, shouldn't there still be some warning to tell people to move away from this variable?

henryiii: We could add that, at the moment I'm just trying to get full compatibility with scikit-build (classic). I don't think it particularly hurts to have False here, though.

PR #1288, opened 2026-04-30
tests: fix Windows PermissionError from concurrent wheelhouse lock races

henryiii:
Debugging an issue gave me this, from copilot:

:robot:

Two separate file locks (main.lock + wheels.lock) allowed concurrent
xdist workers to race on the shared wheelhouse directory: one worker
holding main_lock could call replace()/unlink() while another holding
wheels_lock called zipfile.is_zipfile() on the same files, producing
PermissionError (WinError 5) on Windows.

Fix: consolidate to a single 'wheels.lock' for all wheelhouse mutations.
~~Also avoid redundant pip-wheel rebuilds by checking for the current
scikit-build-core version in the wheelhouse before building (wheel
filenames normalize '+' in the local version label to '_').~~

:rogot: Assisted-by: Copilot:claude-sonnet-4.6

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -76,7 +76,7 @@ def pep518_wheelhouse(
         # already present; this avoids a redundant pip-wheel invocation on every
         # worker while still catching version changes between runs.
         # Wheel filenames normalize the local version label by replacing '+' with '_'.
-        skbuild_version = metadata.version("scikit-build-core").replace("+", "_")
+        skbuild_version = metadata.version("scikit-build-core")
         if not any(
             whl.name.startswith(f"scikit_build_core-{skbuild_version}-")
             for whl in wheelhouse.glob("scikit_build_core-*.whl")
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -67,9 +67,20 @@ def pep518_wheelhouse(
     wheelhouse = pytestconfig.cache.mkdir("wheelhouse")
     tmp_path = tmp_path_factory.mktemp("wheelhouse_tmp")
 
-    main_lock = FileLock(wheelhouse / "main.lock")
-    with main_lock:
-        if not list(tmp_path.glob("scikit_build_core-*.whl")):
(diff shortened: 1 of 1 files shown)


tests/conftest.py
```diff
@@ -67,9 +67,20 @@ def pep518_wheelhouse(
...
+    # A single lock for all wheelhouse mutations prevents cross-lock races on
+    # Windows where concurrent readers (zipfile.is_zipfile) and writers
+    # (unlink/replace) on different locks would cause PermissionError (WinError 5).
+    wheelhouse_lock = FileLock(wheelhouse / "wheels.lock")
+    with wheelhouse_lock:
+        # Only build the scikit-build-core wheel when the current version is not
+        # already present; this avoids a redundant pip-wheel invocation on every
+        # worker while still catching version changes between runs.
+        # Wheel filenames normalize the local version label by replacing '+' with '_'.
+        skbuild_version = metadata.version("scikit-build-core").replace("+", "_")
```
henryiii:
```suggestion
        skbuild_version = metadata.version("scikit-build-core")
```

This wasn't correct.

tests/conftest.py
```diff
@@ -67,9 +67,20 @@ def pep518_wheelhouse(
...
-        if not list(tmp_path.glob("scikit_build_core-*.whl")):
+    # A single lock for all wheelhouse mutations prevents cross-lock races on
+    # Windows where concurrent readers (zipfile.is_zipfile) and writers
+    # (unlink/replace) on different locks would cause PermissionError (WinError 5).
+    wheelhouse_lock = FileLock(wheelhouse / "wheels.lock")
+    with wheelhouse_lock:
+        # Only build the scikit-build-core wheel when the current version is not
+        # already present; this avoids a redundant pip-wheel invocation on every
+        # worker while still catching version changes between runs.
+        # Wheel filenames normalize the local version label by replacing '+' with '_'.
```
henryiii:
```suggestion
```

PR #1292, opened 2026-05-01
fix(editable): recognize Cython `__init__.pxd`/`.pyx` as package indicators

### LINKED ISSUE #1291: Editable installs: packages with `__init__.pxd`/`__init__.pyx` not recognized as packages, causing `__path__` pollution
### Description

The `ScikitBuildRedirectingFinder` in editable installs fails to recognize Cython packages whose `__init__` file has a `.pxd` or `.pyx` extension (rather than `.py`). This causes the parent package's `__path__` to be polluted with child package directories, leading to broken imports.

### Reproducer

This affects any Cython project where:
1. A package directory has both `__init__.py` and `__init__.pxd` (common in Cython projects that export type declarations)
2. The file scanner picks up `__init__.pxd` as the representative file for the package init (due to dict key collision in `mapping_to_modules`)

After an editable install (`pip install -e .`), imports from affected subpackages fail because the wrong `__path__` causes Python's `PathFinder` to resolve the wrong module.

### Root Cause

Three related issues in the editable install machinery:

**1. `_editable_redirect.py` line 113 — Package detection uses substring match on `"__init__.py"`:**

```python
if "__init__.py" in file:
```

This fails for `__init__.pxd` (the substring `"__init__.py"` is not present in `"__init__.pxd"`). As a result, the package is not added to `self.pkgs`, and its directory is added to the **parent**'s `submodule_search_locations` instead of its own.

**2. `_editable_redirect.py` lines 154/163 — `find_spec` gates `submodule_search_locations` on file extension:**

```python
submodule_search_locations=submodule_search_locations
if redir.endswith(("__init__.py", "__init__.pyc"))
else None,
```

Even if issue 1 were fixed, `find_spec` would still refuse to pass `submodule_search_locations` for `.pxd`/`.pyx` init files.

**3. `_editable.py` `mapping_to_modules` — Last-writer-wins collision:**

```python
return {
    path_to_module(Path(v).relative_to(libdir)): str(Path(k).resolve())
    for k, v in mapping.items()
    if is_valid_module(Path(v).relative_to(libdir))
}
```

When both `__init__.py` and `__init__.pxd` exist, `path_to_module` maps both to the same module key. The dict comprehension has no priority logic, so whichever file the filesystem yields last wins — and that may be the `.pxd`, which overwrites the `.py`.

vyasr:
Fixes #1291.

Three targeted fixes for editable installs of Cython projects where packages have `__init__.pxd` or `__init__.pyx` files alongside (or instead of) `__init__.py`:

**`_editable_redirect.py` — package detection (line 113):**
`"__init__.py" in file` → `os.path.basename(file).partition(".")[0] == "__init__"` — matches any `__init__.*` extension, consistent with `path_to_module()` which already handles this correctly.

**`_editable_redirect.py` — `find_spec` (lines 154/163):**
`redir.endswith(("__init__.py", "__init__.pyc"))` → `fullname in self.pkgs` — uses the already-computed package list rather than re-checking file extensions.

**`_editable.py` — `mapping_to_modules`:**
Dict comprehension → explicit loop that prefers `.py`/`.pyc` when multiple source files (e.g. `__init__.py` + `__init__.pxd`) map to the same module key.

### DIFF
--- a/src/scikit_build_core/build/_editable.py
+++ b/src/scikit_build_core/build/_editable.py
@@ -58,11 +58,16 @@ def mapping_to_modules(mapping: dict[str, str], libdir: Path) -> dict[str, str]:
     """
     Convert a mapping of files to modules to a mapping of modules to installed files.
     """
-    return {
-        path_to_module(Path(v).relative_to(libdir)): str(Path(k).resolve())
-        for k, v in mapping.items()
-        if is_valid_module(Path(v).relative_to(libdir))
-    }
+    result: dict[str, str] = {}
+    for k, v in mapping.items():
+        rel = Path(v).relative_to(libdir)
+        if not is_valid_module(rel):
+            continue
+        module = path_to_module(rel)
+        # Prefer .py/.pyc over other extensions (e.g. .pxd, .pyx) for the same module
+        if module not in result or rel.suffix in (".py", ".pyc"):
+            result[module] = str(Path(k).resolve())
+    return result
 
 
 def libdir_to_installed(libdir: Path) -> dict[str, str]:
--- a/src/scikit_build_core/resources/_editable_redirect.py
+++ b/src/scikit_build_core/resources/_editable_redirect.py
@@ -110,7 +110,7 @@ class ScikitBuildRedirectingFinder(importlib.abc.MetaPathFinder):
                 # Strip the last element of the module
                 parent = ".".join(module.split(".")[:-1])
                 # Check if it is a package
-                if "__init__.py" in file:
+                if os.path.basename(file).partition(".")[0] == "__init__":
                     parent = module
                     pkgs.append(parent)
                 # Skip if it's a root module (there are no search paths for these)
@@ -151,7 +151,7 @@ class ScikitBuildRedirectingFinder(importlib.abc.MetaPathFinder):
                 fullname,
                 os.path.join(self.dir, redir),
                 submodule_search_locations=submodule_search_locations
-                if redir.endswith(("__init__.py", "__init__.pyc"))
+                if fullname in self.pkgs
                 else None,
             )
         if fullname in self.known_source_files:
@@ -160,7 +160,7 @@ class ScikitBuildRedirectingFinder(importlib.abc.MetaPathFinder):
                 fullname,
                 redir,
                 submodule_search_locations=submodule_search_locations
-                if redir.endswith(("__init__.py", "__init__.pyc"))
+                if fullname in self.pkgs
                 else None,
             )
         return None
--- a/src/scikit_build_core/resources/_editable_redirect.py
+++ b/src/scikit_build_core/resources/_editable_redirect.py
@@ -128,7 +128,7 @@ class ScikitBuildRedirectingFinder(importlib.abc.MetaPathFinder):
                     parent_path = os.path.join(self.dir, parent_path)
                 submodule_search_locations[parent].add(parent_path)
         self.submodule_search_locations = submodule_search_locations
-        self.pkgs = pkgs
+        self.pkgs = frozenset(pkgs)
 
     def find_spec(
         self,
--- a/tests/test_editable_redirect.py
+++ b/tests/test_editable_redirect.py
@@ -67,7 +67,7 @@ def test_editable_redirect():
             },
         }
     )
-    assert finder.pkgs == ["pkg", "pkg.subpkg"]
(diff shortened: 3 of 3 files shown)


src/scikit_build_core/resources/_editable_redirect.py
```diff
@@ -151,7 +151,7 @@ def find_spec(
                 fullname,
                 os.path.join(self.dir, redir),
                 submodule_search_locations=submodule_search_locations
-                if redir.endswith(("__init__.py", "__init__.pyc"))
+                if fullname in self.pkgs
```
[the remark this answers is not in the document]
henryiii: Seems reasonable, linear checks can become slow. @vyasr?

src/scikit_build_core/build/_editable.py
```diff
@@ -58,11 +58,16 @@ def mapping_to_modules(mapping: dict[str, str], libdir: Path) -> dict[str, str]:
...
-    }
+    result: dict[str, str] = {}
+    for k, v in mapping.items():
+        rel = Path(v).relative_to(libdir)
+        if not is_valid_module(rel):
+            continue
+        module = path_to_module(rel)
+        # Prefer .py/.pyc over other extensions (e.g. .pxd, .pyx) for the same module
+        if module not in result or rel.suffix in (".py", ".pyc"):
+            result[module] = str(Path(k).resolve())
```
LecrisUT, line 69: The logic here is a bit hard to parse, although the whole logic in these is hard to parse and could use a refactor. Iiuc `mapping_to_modules` is used on the src files, and `libdir_to_installed` is used for the CMake installed files. Why not drop all files that do not end with `.py`? It will interfere with `importlib.resources.files`, but I do not think the status-quo is helping us there either.

My proposal is to split (and rename) the inputs to `_make_editable` into 3 variables:
- `source_modules`: mapping of python modules to source `.py` files. All other file types are ignored and presumed that they are compiled. Not sure how to handle `.pyc` though
- `installed_modules`: mapping of python modules to compiled python modules or installed `.py` files
- `file_tree`: a complete file tree of python navigable sources via `importlib.resources.files`. It could be a mapping of modules and list of file paths

@henryiii @vyasr what do you think? The second variable at the very least would unblock me for https://github.com/scikit-build/scikit-build-core/pull/1170

src/scikit_build_core/resources/_editable_redirect.py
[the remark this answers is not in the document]
vyasr: Agreed, storing a temporary set for this would be a simple optimization.

src/scikit_build_core/build/_editable.py
vyasr: > Iiuc mapping_to_modules is used on the src files, and libdir_to_installed is used for the CMake installed files.

Yes this matches my recollection as well.

The split sounds reasonable. I assume `file_tree` will have to be modified substantially as we rework `importlib.resources.files` over time.

src/scikit_build_core/resources/_editable_redirect.py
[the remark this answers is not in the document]
vyasr: Addressed in 4e83d01 — `self.pkgs` is now a `frozenset` for O(1) membership checks in `find_spec`.

src/scikit_build_core/build/_editable.py
vyasr: > Not sure how to handle .pyc though

How many realistic use cases are there for shipping pyc files in wheels? They shouldn't be there _most_ of the time, but I'm sure there are edge cases I'm not thinking of.

LecrisUT: > How many realistic use cases are there for shipping pyc files in wheels? 

For wheels yes it does not make sense, but this part is about usage of editable install paths which is a bit more questionable. I don't know whose responsibility these are or if there is some magic to be careful about in the `importlib`

henryiii: Should we merge this as a fix, then work on a refactor? I could probably just feed that comment into Kimi or GPT and have it do the refactor, actually.

LecrisUT: :+1: for me. LGTM in the current state as an intermediary.

> Not sure how to handle `.pyc` though

will be tricky but we can move that discussion

vyasr: Yeah that seems fine to me.

henryiii (approved):

PR #1297, opened 2026-05-05
feat: add PEP 829 support for editable installs

### LINKED ISSUE #1296: Support PEP 829
PEP 829 replaces import lines in `.pth`.

<details><summary>:robot: Plan summary based on PEP 829</summary>
<p>

Problem

Implement PEP 829 support for scikit-build-core's editable install startup mechanism. Today redirect editable wheels write `_<name>_editable.py` plus `_<name>_editable.pth` containing an `import _<name>_editable` line; inplace editable writes only a path-only `.pth`. PEP 829 changes startup code execution to paired `.start` files on Python 3.15+, while keeping import lines in `.pth` for path extensions and for straddling older Pythons.

Current state

- Redirect editable startup artifacts are produced in `src/scikit_build_core/build/wheel.py::_make_editable`.
- The callable bootstrap implementation already exists in `src/scikit_build_core/resources/_editable_redirect.py`; today it is invoked through `import _<name>_editable` from the `.pth`.
- Redirect editable wheel coverage exists in `tests/test_pyproject_pep660.py`; low-level redirect content coverage exists in `tests/test_editable_unit.py` and `tests/test_editable_redirect.py`.
- Inplace editable mode only writes package paths into `_<name>_editable.pth`, so PEP 829 startup-entry changes should only affect redirect mode.
- Stable ABI / cross-version intent is represented through `wheel.py-api` (for example `cp39`, `cp315t`, `py3`) and tag computation in `builder/wheel_tag.py`.

Proposed approach

1. Add a small helper in the editable wheel-writing path that decides which startup artifacts to emit for redirect editable wheels:
   - always keep `_<name>_editable.py`
   - keep `_<name>_editable.pth` import statements for older Python compatibility and Cython path injection
   - add `_<name>_editable.start` pointing at a zero-argument callable inside `_<name>_editable.py`
   - gate `.start` emission so it is always written on Python 3.15+, and also written on older builders for non-version-specific tags (Stable ABI `cp3x` / `cp3xxt` and Pythonless `py3` / `py2.py3`) because those wheels may later be installed on Python 3.15+
2. Refactor the generated editable bootstrap so the startup action can be referenced as a PEP 829 entry point such as `_<name>_editable:install`.
3. Keep inplace editable behavior unchanged unless investigation during implementation shows a compatibility reason to touch it.
4. Extend tests around editable wheel contents and generated bootstrap text to cover:
   - redirect wheels on Python 3.15+ producing `.start`
   - redirect wheels on Python < 3.15 keeping compatibility behavior
   - Stable ABI / non-version-specific cases that must still emit `.start`
   - inplace wheels continuing to avoid executable startup artifacts
5. Update user-facing docs if editable wheel artifact behavior or compatibility guidance is documented.

Notes / considerations

- PEP 829 requires `.start` files to contain `pkg.mod:callable` entries, so the generated redirect module likely needs a dedicated zero-argument entry-point wrapper instead of relying on side effects at import time.
- The existing `.pth` also adds package source paths for Cython support; the plan should preserve those path lines regardless of `.start` handling.
- Scope decision: treat Pythonless tags like Stable ABI tag

</p>
</details>

henryiii:
Fix #1296. (Now reworked)

### DIFF
--- a/src/scikit_build_core/build/_editable.py
+++ b/src/scikit_build_core/build/_editable.py
@@ -1,6 +1,7 @@
 from __future__ import annotations
 
 import os
+import sys
 import typing
 from collections.abc import Mapping
 from pathlib import Path
@@ -41,9 +42,14 @@ def editable_redirect(
     build_options: Sequence[str],
     install_options: Sequence[str],
     install_dir: str,
+    as_entrypoint: bool = False,
 ) -> str:
     """
     Prepare the contents of the _editable_redirect.py file.
+
+    If ``as_entrypoint`` is set, the install call is wrapped in a zero-argument
+    ``entrypoint()`` function (for PEP 829 ``.start`` files) rather than being
+    invoked at module import time (for the legacy ``.pth`` ``import`` line).
     """
 
     editable_py = resources / "_editable_redirect.py"
@@ -60,7 +66,10 @@ def editable_redirect(
         install_dir,
     )
     arguments_str = ", ".join(repr(x) for x in arguments)
-    editable_txt += f"\n\ninstall({arguments_str})\n"
+    if as_entrypoint:
+        editable_txt += f"\n\ndef entrypoint() -> None:\n    install({arguments_str})\n"
+    else:
+        editable_txt += f"\n\ninstall({arguments_str})\n"
     return editable_txt
 
 
@@ -74,7 +83,19 @@ def editable_redirect_files(
     packages: Iterable[str],
     reload_dir: Path | None,
     settings: ScikitBuildSettings,
+    use_start: bool | None = None,
 ) -> dict[str, bytes]:
+    """
+    Build the editable redirect files for a package.
+
+    On Python 3.15+ (PEP 829), the ``import`` line that runs the redirect is
+    moved out of the ``.pth`` file (where it is deprecated) into a ``.start``
+    file, and the ``.pth`` keeps only the ``sys.path`` entries. ``use_start``
+    overrides this auto-detection (used by tests); leave it ``None`` to select
+    based on the running interpreter.
+    """
+    if use_start is None:
+        use_start = sys.version_info >= (3, 15)
     modules = mapping_to_modules(mapping, libdir)
     installed = libdir_to_installed(libdir)
     if settings.wheel.install_dir.startswith("/"):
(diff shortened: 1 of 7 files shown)


src/scikit_build_core/build/_editable.py
```diff
@@ -90,12 +114,17 @@ def editable_redirect_files(
...
     )
-    package_paths = tuple(packages)
-    pth_import_paths = "\n".join([f"import _{name}_editable", *package_paths, ""])
-    return {
+    editable_files = {
         f"_{name}_editable.py": editable_txt.encode(),
-        f"_{name}_editable.pth": pth_import_paths.encode(),
+        f"_{name}_editable.pth": editable_redirect_pth(
+            name=name, packages=packages
+        ).encode(),
```
LecrisUT: Should we make this conditional on python version? Would like the PEP829 format to be more guaranteed used

src/scikit_build_core/build/_editable.py
```diff
@@ -18,15 +18,21 @@
...
     "editable_redirect_files",
+    "editable_redirect_pth",
+    "editable_redirect_start",
     "get_packages",
     "libdir_to_installed",
     "mapping_to_modules",
 ]
 
 
+EDITABLE_REDIRECT_CALLABLE = "bootstrap"
```
LecrisUT: I don't find a reference to this

src/scikit_build_core/build/wheel.py
```diff
@@ -56,19 +64,49 @@ def _make_editable(
...
+
+    wheel.writestr(
+        f"_{name}_editable.py",
+        editable_txt.encode(),
+    )
+    wheel.writestr(
+        f"_{name}_editable.pth",
+        editable_redirect_pth(name=name, packages=packages).encode(),
+    )
+    if _should_write_editable_start(wheel_tags):
```
LecrisUT: I don't think we need this complication. This is not being packaged or anything so wr only care what the local python version is

src/scikit_build_core/build/_editable.py
```diff
@@ -60,10 +66,27 @@ def editable_redirect(
...
+    import_strings = [
+        f"import _{name}_editable; _{name}_editable.{EDITABLE_REDIRECT_CALLABLE}()",
+        *packages,
+        "",
+    ]
+    return "\n".join(import_strings)
+
+
+def editable_redirect_start(*, name: str) -> str:
+    return f"_{name}_editable:{EDITABLE_REDIRECT_CALLABLE}\n"
```
LecrisUT: From my understanding of this format, we must not do the additional `install()` command inside the `_editable_redirect.py`, instead it is handled by `EDITABLE_REDIRECT_CALLABLE`. It seems hard to satisfy both fotmats simultaneously, but I don't think we need to do that

src/scikit_build_core/build/_editable.py
LecrisUT: Nvm found it

src/scikit_build_core/build/_editable.py
LecrisUT: Ah it's literally a few lines above. I should deffer reviews when I'm at the PC

src/scikit_build_core/build/_editable.py
LecrisUT: Hmm, according yo the pep, the start should have priority. At least could embed thr higher bound when it is expected to be dropped so we don't forget about it

src/scikit_build_core/build/wheel.py
henryiii: 🤦 I was really overcompensating this thinking it was distributed.

Issue #1299, opened 2026-05-10
AI review (take with grain of salt!)


henryiii:
I'm busy at PyCon, dropping this here. From Kimi-K2.6:

:robot:

Here's my comprehensive review of the repository, organized by category:

---

## 🔴 Bugs / Logic Errors

| File | Issue | Details |
|------|-------|---------|
| `src/scikit_build_core/build/wheel.py:216` | `.format()` on user-controlled string | `settings_reader.settings.messages.after_failure.format()` — if the message contains `{` or `}`, it raises `KeyError`/`IndexError`, masking the original error. **Fix:** remove `.format()` |
| `src/scikit_build_core/builder/builder.py:234,241` | Missing format argument in logger | `logger.info("py-api {} requires CPython, ignoring")` — `{}` has no argument; raises `IndexError` on PyPy. **Fix:** pass `py_api` |
| `src/scikit_build_core/cmake.py:90` | Mutable class-level default | `single_config: bool = not sysconfig.get_platform().startswith("win")` evaluated once at import time; stale after mocks. **Fix:** use `field(init=False)` + `__post_init__` |
| `src/scikit_build_core/settings/skbuild_read_settings.py:220-222` | Unsafe dict access | `pyproject["build-system"]["requires"]` crashes with `KeyError` if `[build-system]` is missing. **Fix:** use `.get("build-system")` with helpful error |
| `src/scikit_build_core/settings/sources.py:772,785` | Overly broad `except Exception` | Catches `AttributeError` etc. from programming bugs. **Fix:** narrow to `TypeError, ValueError` |

---

## 🟠 Performance Issues

| File | Issue | Recommendation |
|------|-------|---------------|
| `src/scikit_build_core/build/wheel.py:410-468` | Dead `build_args` variable | Always empty list; remove it |
| `src/scikit_build_core/builder/builder.py:132-145` | Entry points loaded 3× per configure | Cache `_get_entry_point_search_path` with `@lru_cache` |
| `src/scikit_build_core/builder/builder.py:326-328` | `get_archs` recomputed | Already computed in `wheel.py`; pass the result instead |
| `src/scikit_build_core/program_search.py:41-59` | `_macos_binary_is_x86` subprocess overhead | Cache with canonicalized `Path` |
| `src/scikit_build_core/program_search.py:129-173` | Duplicate subprocess for CMake version | Consider defaulting to `--version` if JSON path often fails |
[...]

PR #1305, opened 2026-05-31
fix: minor fixes based on review

henryiii:
This is some of the useful things from #1299, worked through all sections except the first. Many comments there were false positives, but there were a few good ones.

Also adding `.claude` because ClaudeCode sticks stuff there.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -178,3 +178,5 @@ uv.lock
 
 # Symlink to AGENTS.md
 CLAUDE.md
+# ClaudeCode sticks stuff in here
+.claude/
--- a/.gitignore
+++ b/.gitignore
@@ -178,5 +178,5 @@ uv.lock
 
 # Symlink to AGENTS.md
 CLAUDE.md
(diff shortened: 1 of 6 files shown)


.gitignore
```diff
@@ -178,3 +178,5 @@ uv.lock
 
 # Symlink to AGENTS.md
 CLAUDE.md
+# ClaudeCode sticks stuff in here
```
henryiii:
```suggestion
# Claude Code sticks stuff in here
```

src/scikit_build_core/build/wheel.py
```diff
@@ -412,6 +412,9 @@ def _build_wheel_impl_impl(
                 "{green}***",
                 f"{{bold}}Building project with {{blue}}{generator}{{default}}...",
             )
+
+            # These are the args _before_ the `--`, directly to `--build`
```
henryiii:
```suggestion
            # These are the args before the `--`, directly to `--build`
```

Issue #1298, opened 2026-05-05
7 tests fail


yurivict:
[log](https://freebsd.org/~yuri/scikit-build-core-0.12.2-tests.log)

Version: 0.12.2

henryiii:
<details>
<summary>:robot: Claude's analysis:</summary>

The one genuine bug: `test_requires_command[False-sdist]`

This is a test-isolation bug in our suite. rich_warning is memoized with @functools.lru_cache(maxsize=None) (_logging.py:341), so the "CMakeLists.txt not found…" warning is emitted at most once per process. The test calls rich_warning.cache_clear() after main() instead of before:

```
main()
rich_warning.cache_clear()   # too late — clears for next time, not this time
out, err = capsys.readouterr()
assert "CMakeLists.txt not found" in err
```

Under xdist, if any earlier test in the same worker emits that exact warning, this test's main() hits the cache and prints nothing → err == ''. I reproduced it above. It's distribution-dependent, which is why only [False-sdist] failed for them and it passes for us. Fix: clear before main().

The other 6 are environment issues, not scikit-build-core bugs

- test_pep517_wheel (Fortran) — NumPy 1.x/2.4.4 ABI mismatch in their build env (module compiled against NumPy 1.x, run under 2.4.4). Their packaging, not ours.
- test_abi3_wheel, test_pep517_wheel[simple_setuptools_ext], test_toml_wheel[...], test_mixed_wheel[...] — all the same FutureWarning about the standalone wheel package no longer providing bdist_wheel, turned into an error by filterwarnings=error. They have an old setuptools (<70.1) / stale wheel. Environment.
- test_valid_schemas_files[cmake_defines] — our schema is correct (verified at the v0.12.2 tag: it has the string|bool|array branch). The oneOf-inside-oneOf is redundant, and their validate_pyproject/fastjsonschema version evaluates the nested form as "0 matches". Flattening the nested oneOf would make it more robust across validator versions, but it's not wrong as-is.
- 16 collection errors — ValueError: invalid pyproject.toml config: 'project.license' from a newer setuptools rejecting the sample setuptools packages' license format. Environment (their setuptools version).

So: 1 real (test-only) bug, the rest are their NumPy/setuptools/validate_pyproject versions.

</details>

I've made a PR fixing some of those.

PR #1306, opened 2026-05-31
fix: test isolation and schema/warning robustness (#1298)

henryiii:
:robot:

Addresses the test failures reported in #1298. Of the 7 failures + 16 errors, most are environment-specific on the reporter's FreeBSD box (NumPy 1.x/2.x ABI mismatch, newer setuptools rejecting `project.license`), but a few are genuine robustness issues fixed here.

## Changes

- **`tests/test_build_cli.py`** — `rich_warning` is `lru_cache`-memoized, so the `"CMakeLists.txt not found"` warning is emitted at most once per process. `test_requires_command` cleared the cache *after* `main()`, so under xdist an earlier test in the same worker could suppress the warning, leaving `err == ''` (the reporter's `assert 'CMakeLists.txt not found' in ''`). Now cleared *before* `main()`.

- **`src/scikit_build_core/settings/json_schema.py`** — flatten the `oneOf`-in-`oneOf` generated for `EnvVar`-annotated `Union` fields (e.g. `cmake.define`). The branches are mutually exclusive by JSON type (string/bool/array/object), so the flat `oneOf` is semantically identical but avoids a `validate_pyproject`/`fastjsonschema` version quirk that reported "0 matches found" for `cmake.define` lists. Schema regenerated via `nox -t gen`.

- **`pyproject.toml`** — add the `FutureWarning` variant of the existing `"The 'wheel' package is no longer the canonical location"` filter. The old filter only matched `DeprecationWarning`; newer wheel/setuptools raise it as a `FutureWarning`, which `filterwarnings=error` promoted to a failure in the setuptools tests.

## Not addressed (reporter's environment, not fixable here)

- Fortran `test_pep517_wheel` — NumPy 1.x/2.4.4 ABI mismatch in the build env.
- 16 collection errors — `invalid pyproject.toml config: 'project.license'` from a newer setuptools rejecting the sample packages' license format.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -184,6 +184,7 @@ filterwarnings = [
     "ignore:Config variable '.*' is unset, Python ABI tag may be incorrect:RuntimeWarning",
     "default:pkg_resources is deprecated as an API:DeprecationWarning:wheel",  # Caused by wheel<0.41 in tests
     "default:The 'wheel' package is no longer the canonical location:DeprecationWarning",  # Caused by wheel also
+    "default:The 'wheel' package is no longer the canonical location:FutureWarning",  # Newer wheel/setuptools emit this as a FutureWarning
     "default:onerror argument is deprecated, use onexc instead:DeprecationWarning:wheel", # Caused by wheel<0.41 & Python 3.12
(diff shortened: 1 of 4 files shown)


LecrisUT (approved):
Not sure what part is supposed to address test isolation. But otherwise lgtm

PR #1308, opened 2026-06-01
docs: avoid a common timeout

henryiii:
This is timing out sometimes, let's just assume the learn site is up. Also touched up a few things that report redirects.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -222,7 +222,7 @@ jobs:
     name: Manylinux on 🐍 3.14 • Free-threaded
     runs-on: ubuntu-latest
     timeout-minutes: 40
-    container: quay.io/pypa/musllinux_1_2_x86_64@sha256:36bd94f44a613d006b5255f8393ab262d86a3bc016feedb91dc0b520a1b51e50 # 2026.05.17-1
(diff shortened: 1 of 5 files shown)


LecrisUT:
If you are interested, there is a PR for caching linkcheck results [^1], and there is a workaround to backport it to a local extension [^2]

[^1]: https://github.com/sphinx-doc/sphinx/pull/14424
[^2]: [sphinx extension](https://github.com/teemtee/tmt/blob/main/docs/ext/linkcheck_cache.py) and [github action](https://github.com/teemtee/tmt/blob/4ba84dc49ff30ed82435a02635cd048a9630410f/.github/workflows/step-doc-tests.yml#L52-L59)

henryiii:
Nice, might be something to try in the future. The problem currently with GitHub.com links is that we get rate limited in a single run, that would be hard to cache. I manually ran some of them by commenting out the exclusion (which is how I found the Microsoft -> lightgdm redirect), and got rate limited near the end of the run. Manually listing some that we know work in the config file would be similar to a cache.

LecrisUT:
> we get rate limited in a single run, that would be hard to cache

Yeah it's a hard problem. The caching is done on a per-link basis, so in theory if the link checks are distributed it can still work. But in the workaround the upload is an all-or-nothing action and would be hard to "seed" the cache. The PR would solve that issue though.

PR #1310, opened 2026-06-01
fix: more reliably find the Python library on Windows

henryiii:
This is based on the CMake search procedure, but in Python.

:robot: _AI text follows_ :robot:

## Problem

`get_python_library` located the Python library purely from POSIX Makefile config vars (`LIBDIR`, `LDLIBRARY`, `LIBRARY`, `MULTIARCH`, `PYTHONFRAMEWORKPREFIX`). On **Windows these are almost always `None`**, so the function returned `None` — exactly where it matters most. Windows extension/embedding builds must link against `pythonXY.lib`; there is no ELF lazy symbol resolution to paper over a missing library. As a result `{prefix}_LIBRARY` was effectively never set on Windows, leaving FindPython to guess (which fails for venvs, conda, embeddable/nuget layouts, and cross-compiles).

## Approach

Rewrite the locator to mirror CMake's FindPython (`Support.cmake`): construct the library *name* and search a known directory layout under the install prefix, rather than relying on Unix-only Makefile vars.

- **Windows**: probe `<base_exec_prefix>/{libs,lib}` for `python3<minor>[t][_d].lib` (or the stable-ABI `python3[t].lib`). `base_exec_prefix` (not `prefix`) ensures venvs and conda envs resolve to the real `libs/` dir. `_d` debug suffix and `t` free-threaded flag mirror CMake's `_PYTHON_GET_NAMES`.
- **POSIX/macOS**: existing `LDLIBRARY`/`LIBRARY` + multiarch + framework logic, restructured into a name×directory search, with additive existence-checked `LIBPL` (CMake's CONFIGDIR) and `lib64` fallbacks. Still returns `None` when no real file is found (a wrong hint can break FindPython, and ELF lazy-binding makes it unnecessary).
- The `DIST_EXTRA_CONFIG` cross-compile branch is preserved byte-for-byte.

Everything is existence-gated, so any case that returned a value before still does. `builder.py` keeps the Windows-only `{prefix}_LIBRARY` guard — it just starts firing now that the locator returns real paths on Windows.

## Tests

- `test_get_python_library`: dropped the `xfail`; Windows assertions are now real (name recomputed independently of the implementation, parent in `{libs,lib}`, plus abi3/abi3t).
- New `test_windows_lib_names`: cross-platform parametrized matrix {classic, abi3, abi3t} × {gil} × {debug}, no filesystem access, runs on all CI OSes.
- The cross-compile `test_get_python_library_xcompile` is unchanged.

`prek -a --quiet` (ruff + strict mypy) passes; `pytest tests/test_builder.py` green locally on macOS. The real `.is_file()` Windows assertions only exercise on the Windows CI job.

### DIFF
--- a/src/scikit_build_core/builder/builder.py
+++ b/src/scikit_build_core/builder/builder.py
@@ -283,7 +283,9 @@ class Builder:
                 cache_config[f"{prefix}_ROOT_DIR"] = Path(sys.base_exec_prefix)
                 cache_config[f"{prefix}_INCLUDE_DIR"] = python_include_dir
                 cache_config[f"{prefix}_FIND_REGISTRY"] = "NEVER"
-                # FindPython may break if this is set - only useful on Windows
+                # On Windows the library is constructed and existence-checked, so
+                # this is reliable. On POSIX it can break FindPython (and the
+                # locator returns None there anyway), so it stays Windows-only.
                 if python_library and sysconfig.get_platform().startswith("win"):
                     cache_config[f"{prefix}_LIBRARY"] = python_library
                 if python_sabi_library and sysconfig.get_platform().startswith("win"):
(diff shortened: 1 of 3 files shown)


henryiii:
Had Fable look for fixes, then asked Opus to trim it down, remove 80% of the diff. This is focused on Windows now, so I'm pretty happy with it.

PR #1312, opened 2026-06-01
refactor: simplification cleanups from a repo sweep

henryiii:
:robot: Non-behavioral cleanups found alongside #1311 (the bug-fix sweep). Each is independent and low-risk; no functional change intended.

### Cleanups

1. **`settings/sources.py`** — removed a redundant `is_union_type(...) and str in get_args(target)` early-return. The following union block re-checks `is_union_type` and returns `item` for the same `str` case, so the early return was dead.

2. **`builder/builder.py`** — removed a redundant `target_minor_version = int(py_api[3:])` in the classic-abi3 branch. It was already computed identically a few lines up (`int(py_api[3:].rstrip("t"))`, and the trailing-`t` case is handled by an earlier `elif`).

3. **`builder/generator.py`** — moved the `cmake --help` subprocess after the non-MSVC-Windows early return, since on MinGW/MSYS the result is discarded (always `Ninja`). Avoids a wasted subprocess there.

4. **`builder/sysconfig.py`** — the "Detected ABI3 SOABI" / "ABI3T SOABI" lines used a `{color}` prefix instead of `{bold}`, so they rendered without the bold styling every other line in `info_print` gets. Cosmetic.

5. **`cmake.py`** — only emit the uppercase `<PKG>_ROOT` cache entry when it differs from the original name, avoiding an identical duplicate `set(...)` line when the package name is already uppercase.

6. **`file_api/reply.py`** — `model.directory.Directory` was in the `make_class` auto-resolve set but has no `jsonFile` field (the class with one is `codemodel.Directory`, intentionally not resolved), so the entry never fired. Removed it and its now-unused import.

### Verification

- `prek -a` clean (ruff + mypy)
- `test_fileapi`, `test_skbuild_settings`, `test_builder`, `test_generator_default` all pass

(Note: `test_cmake_config` has pre-existing failures on `main` in this environment — an unregistered `lipo -info` call under pytest_subprocess — unrelated to these changes.)

### DIFF
--- a/src/scikit_build_core/builder/builder.py
+++ b/src/scikit_build_core/builder/builder.py
@@ -242,15 +242,13 @@ class Builder:
                         py_api,
                         target_minor_version,
                     )
-            else:
-                # Classic stable ABI (abi3)
-                target_minor_version = int(py_api[3:])
-                if gil_disabled:
-                    logger.info(
-                        "Free-threaded Python doesn't support the classic Limited API, ignoring"
-                    )
-                elif target_minor_version <= sys.version_info.minor:
-                    sabi = _SabiMode.ABI3
+            # Classic stable ABI (abi3)
+            elif gil_disabled:
+                logger.info(
+                    "Free-threaded Python doesn't support the classic Limited API, ignoring"
+                )
+            elif target_minor_version <= sys.version_info.minor:
+                sabi = _SabiMode.ABI3
 
         python_library = get_python_library(self.config.env, abi3=False)
         python_sabi_library = None
--- a/src/scikit_build_core/builder/generator.py
+++ b/src/scikit_build_core/builder/generator.py
@@ -60,12 +60,12 @@ def get_default(cmake: CMake) -> str | None:
     """
     Returns the computed default for the current platform.
     """
-    generator = get_default_from_cmake(cmake)
-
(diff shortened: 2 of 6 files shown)


LecrisUT:
> 6\. **`file_api/reply.py`** — `model.directory.Directory` was in the `make_class` auto-resolve set but has no `jsonFile` field (the class with one is `codemodel.Directory`, intentionally not resolved), so the entry never fired. Removed it and its now-unused import.

Btw this was not correct. I am trying to do a refactor of this though so it should not be an issue in the long term
```json
			"directories" : 
			[
				{
					"build" : ".",
					"hasInstallRule" : true,
					"jsonFile" : "directory-.-5e7a28751b0c9235cbe7.json",
					"minimumCMakeVersion" : 
					{
						"string" : "3.15"
					},
					"projectIndex" : 0,
					"source" : ".",
					"targetIndexes" : 
					[
						0
					]
				}
			],
```
just caught this when refactoring the branch

henryiii:
I believe it was correct that it wasn't working before, it just wasn't the right fix (should have been added fully rather than removed) - this just removed the working code?

LecrisUT:
Yes, the cattrs comparison is what was missing. The non-cattrs actually worked right when I tried it

Issue #1317, opened 2026-06-10
AI review (Claude Fable 5)


henryiii:
:robot: _AI text below_ :robot:

All findings I spot-checked verified in source. Here's the full review.

I ran six parallel review agents over the subsystems (settings, build, builder/cmake, file_api/ast/metadata, hatch/setuptools, top-level utilities), then personally verified the highest-severity claims by reading the code. Several were also reproduced empirically by the agents. Findings below, ordered by severity; I verified everything in the "high-severity" section myself.

## High-severity bugs

**1. `{root}` formats to the dataclass repr** — `format.py:111`. `RootPathResolver.__format__("")` returns `str(self)`, but the class defines no `__str__`, so the dataclass repr leaks out: `"{root}".format(...)` produces `RootPathResolver(path=PosixPath('/...'))` instead of the path. Only `{root:uri}` is covered by tests. Affects `build-dir`, `build.requires`, etc. Fix: `return str(self.path)`.

**2. Editable-rebuild errors print `ERROR: None` or vanish** — `resources/_editable_redirect.py:398` and `:420`. The condition is inverted: when `verbose`, stdout went straight to stderr so `result.stdout` is `None` (prints `ERROR: None`); when not verbose (the default), the captured output is never printed and `check_returncode()` raises without it — build errors written to stdout (e.g. MSBuild) disappear. Should be `and not verbose`. Both the `--build` and `--install` blocks have it.

**3. Hatch plugin double-applies `wheel.install_dir` for regular wheels** — `hatch/plugin.py:347-348`. CMake already installs into `wheel_dirs[targetlib] / settings.wheel.install_dir` (line 224), and the non-editable `force_include` mapping prefixes `install_dir` again, producing e.g. `hatchling_example/hatchling_example/_core.so` — an unimportable wheel. The editable branch maps without the prefix (and its test imports successfully); `test_hatchling_wheel` asserts the doubled path but never imports the wheel, so it encodes the bug.

**4. Override with `inherit` table plus any top-level non-dict key crashes** — `settings/skbuild_overrides.py:461-463`. The non-dict branch does `inherit_override_tmp = inherit_override or "none"` then `assert not inherit_override_tmp` if it's a dict — asserting the *whole* `inherit` table is empty. A schema-valid config combining `inherit.cmake.args = "append"` with a scalar key like `build-dir` dies with a bare `AssertionError` (reproduced by the agent). The else branch should use `"none"` per-key like the dict branch does.

[...]

PR #1320, opened 2026-06-10
fix(builder): cmake generator detection, archs cross-compile, and program search robustness

henryiii:
Autogenerated from #1317.

This PR fixes a reviewed set of bugs in `builder/`, `cmake.py`, and `program_search.py`, with regression tests. Each finding was verified against the source before fixing.

## Fixes

1. **`cmake.py` version-error message** — `"CMake version undetermined @ {program.path}"` was missing the `f` prefix and referenced an out-of-scope variable. Now `f"...{cmake_program.path}"`.

2. **`cmake.py` multi-config generator detection** — `CMaker.configure` called `get_generator(*all_args)` without forwarding `defines`, so a multi-config generator set via `cmake.define.CMAKE_GENERATOR` (e.g. `"Ninja Multi-Config"`, `"Xcode"`) was not detected. `single_config` stayed `True`, `-DCMAKE_BUILD_TYPE` was passed (ignored by multi-config generators) and `--config` was omitted from build/install — silently building the wrong config. Now passes `defines=defines or {}`. New test: `test_cmake_multi_config_generator_via_define`.

3. **`builder/generator.py` Ninja Multi-Config handling** — `set_environment_for_gen` used an exact-match (`(generator or "Ninja") == "Ninja"`), so an explicit `"Ninja Multi-Config"` skipped all ninja handling (no `CMAKE_MAKE_PROGRAM` hint, no `ninja.version` enforcement, no `NinjaNotFoundError`). Now matches any Ninja generator (mirroring `_uses_ninja_generator`) and records the actual generator name in `CMAKE_GENERATOR`. New tests: `test_set_environment_for_gen_ninja_variants`, `test_set_environment_for_gen_ninja_multi_config_missing`.

4. **`builder/builder.py` dead `get_archs` cross-compile path** — `get_archs(env, cmake_args=())` checks `cmake_args` for `CMAKE_SYSTEM_PROCESSOR` (added for conda cross-compiles in #207) and the docs document this, but no caller ever passed `cmake_args`. Wired the configured cmake args through the wheel-tag/arch paths (`Builder.configure`, `build/wheel.py`, `hatch/plugin.py`, `setuptools/build_cmake.py`) via a shared `get_cmake_args_from_settings` helper, so `-DCMAKE_SYSTEM_PROCESSOR` in `cmake.args`/`CMAKE_ARGS` now suppresses `ARCHFLAGS` handling on macOS as documented. All four call sites were safely wired through. New test: `test_builder_macos_arch_cmake_system_processor`. The docs (`docs/guide/crosscompile.md`) already matched the now-implemented behavior, so no doc change was needed.

5. **`builder/__main__.py` "Get Requires" diagnostic** — added the missing `*req.variants()` so the diagnostic matches the real hooks.

6. **`program_search.py` fallback error handling** — if `cmake -E capabilities` raised `CalledProcessError`, the `--version` fallback ran inside that handler; a `TimeoutExpired`/`PermissionError` raised there skipped the sibling `except` clauses and propagated out of `get_cmake_program`, aborting the whole search. Restructured with a nested `try` so those errors are handled the same way, yielding `Program(path, None)`. New test: `test_get_cmake_program_fallback_exception`.

7. **`builder/macos.py` simplification** — `get_macosx_deployment_target` manually re-normalized `env_target`, duplicating `normalize_macos_version`. Now passes the raw target through (behavior preserving aside from log text; covered by existing `test_macos_version`).

## Skipped

None — all seven findings were verified and fixed.

## Tests

```
uv run pytest tests/test_builder.py tests/test_program_search.py tests/test_cmake_config.py tests/test_get_requires.py -q -n auto
# 76 passed, 3 skipped, 1 xfailed
```

`prek -a --quiet` passes (ruff, ruff-format, mypy, etc.). The only failing hook is `check-sdist`, which fails identically on a clean checkout of `main` in this worktree (complains about an untracked `.claude/settings.local.json` and `uv.lock`); it is unrelated to these changes.

### DIFF
--- a/src/scikit_build_core/builder/builder.py
+++ b/src/scikit_build_core/builder/builder.py
@@ -31,7 +31,12 @@ if TYPE_CHECKING:
     from ..cmake import CMaker
     from ..settings.skbuild_model import ScikitBuildSettings
 
-__all__ = ["Builder", "archs_to_tags", "get_archs"]
+__all__ = [
+    "Builder",
+    "archs_to_tags",
+    "get_archs",
+    "get_cmake_args_from_settings",
+]
 
 DIR = Path(__file__).parent.resolve()
 
@@ -92,6 +97,25 @@ def _filter_env_cmake_args(env_cmake_args: list[str]) -> Generator[str, None, No
             yield arg
 
 
+def get_cmake_args_from_settings(
+    settings: ScikitBuildSettings, env: Mapping[str, str]
+) -> list[str]:
+    """
+    Get CMake args from the settings and environment (settings ``cmake.args``
+    plus the filtered ``CMAKE_ARGS`` environment variable).
+    """
+    # Adding CMake arguments set as environment variable
+    # (needed e.g. to build for ARM OSX on conda-forge)
+    env_cmake_args: list[str] = list(
+        filter(None, shlex.split(env.get("CMAKE_ARGS", "")))
+    )
+
+    if env_cmake_args:
+        logger.debug("Env CMAKE_ARGS: {}", env_cmake_args)
+
+    return [*settings.cmake.args, *_filter_env_cmake_args(env_cmake_args)]
+
+
 def _sanitize_path(path: Any) -> list[Path]:
     # This handles classes like:
     # MultiplexedPath from importlib.resources.readers (3.11+)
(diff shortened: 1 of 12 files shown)


src/scikit_build_core/builder/builder.py
```diff
@@ -92,6 +97,25 @@ def _filter_env_cmake_args(env_cmake_args: list[str]) -> Generator[str, None, No
...
+    settings: ScikitBuildSettings, env: Mapping[str, str]
+) -> list[str]:
+    """
+    Get CMake args from the settings and environment (settings ``cmake.args``
+    plus the filtered ``CMAKE_ARGS`` environment variable).
+    """
+    # Adding CMake arguments set as environment variable
+    # (needed e.g. to build for ARM OSX on conda-forge)
+    env_cmake_args: list[str] = list(
+        filter(None, shlex.split(env.get("CMAKE_ARGS", "")))
```
LecrisUT, line 110: Isn't the `filter(None)` a no-op here?

Other than that, this seems like it should be cached?

PR #1322, opened 2026-06-10
fix(build): entry-point, wheel, and requires-state bugs

henryiii:
Autogenerated from #1317.

Fixes a reviewed batch of bugs in `src/scikit_build_core/build/` plus the
`get_requires` state plumbing, with regression tests. Each finding was verified
against the source before fixing.

## Bug fixes

1. **`build/__init__.py` `__all__` duplication / undefined names** — The
   top-level `__all__` unconditionally listed `prepare_metadata_for_build_wheel`
   and `prepare_metadata_for_build_editable`, which are only defined (and
   appended to `__all__`) inside `if _has_safe_metadata():`. This produced
   duplicates when metadata was safe and advertised undefined names (breaking
   `import *`) otherwise. Removed them from the top-level list.

2. **`wheel.py` ignored explicit `license-files = []`** — `if metadata.license_files:`
   used truthiness, so opting out of PEP 639 license files fell into the default
   license-glob branch and shipped files METADATA did not list. Now uses
   `is not None` (matching the validation check above). Regression test:
   `test_pep639_license_files_optout`.

3. **Generated `location = "source"` files written without UTF-8** — both
   `wheel.py` and `sdist.py` wrote source-location generated files without
   `encoding="utf-8"` (unlike the build/install path), breaking non-ASCII output
   on non-UTF-8 locales (Windows). Added `encoding="utf-8"` at both sites.

4. **`_scripts.py` crashed on subdirectories in the scripts dir** —
   `process_script_dir` opened every `iterdir()` entry, suppressing only
   `UnicodeDecodeError`; a subdirectory raised `IsADirectoryError`
   (`PermissionError` on Windows). Now skips non-files. Regression test:
   `test_script_dir_with_subdirectory`.

5. **`_file_processor.py` mutated `dirs` while iterating it** — excluded
   directory names were removed from `dirs` during `for dname in dirs:`,
   skipping the next sibling so consecutive excluded dirs were not all pruned,
   letting `os.walk(followlinks=True)` descend into excluded trees. Now iterates
   `list(dirs)`. Regression test: `test_consecutive_excluded_dirs_not_descended`.

6. **`get_requires_for_build_*` resolved overrides with the wrong state** —
   `GetRequires.from_config_settings` always used the default `state="sdist"`, so
   `if.state = "wheel"`/`"editable"` overrides were ignored when computing build
   requirements for the wheel/editable hooks. Plumbed a `state` argument through
   and pass `"sdist"`/`"wheel"`/`"editable"` from the respective hooks.
   Regression test: `test_get_requires_state_override`.

## Simplifications

7. **`build/__init__.py`** — extracted the thrice-copied
   `FailedLiveProcessError` → flush → `rich_print` → `SystemExit(1)` block into
   an `_exit_on_failed_live_process` context manager, and factored the
   byte-identical wheel/editable requirement bodies into a shared
   `_get_requires_for_build_wheel` helper differing only by state.

8. **`wheel.py`** — factored the two identical `WheelWriter(...)` constructions
   (metadata-only and full-wheel paths, differing only in `folder`) into one
   `functools.partial`.

## Skipped

- **Optional `_file_processor.py` `nested_excludes` caching** — `each_unignored_file`
  runs a full `os.walk(".")` to discover nested `.gitignore`s on every call, and
  `packages_to_file_mapping` calls it once per package. Caching it cleanly would
  require module-level memoization keyed on cwd (the walk always starts at `.`
  regardless of `starting_path`), which is fragile across the test suite's
  `chdir`-based fixtures. Skipped to avoid contorting the API, as the task noted
  this was optional and "only if clean."

## Tests

Targeted suites pass:

```
tests/test_pyproject_pep517.py tests/test_file_processor.py
tests/test_get_requires.py tests/test_process_scripts.py
tests/test_prepare_metadata.py tests/test_settings_overrides.py
→ 152 passed, 1 skipped
```

`prek -a --quiet` passes ruff + mypy on the changes. The only `prek` failure is
`check-sdist`, which flags untracked worktree artifacts (`uv.lock`,
`.claude/settings.local.json`) present in this agent worktree and unrelated to
these changes.

### DIFF
--- a/src/scikit_build_core/build/__init__.py
+++ b/src/scikit_build_core/build/__init__.py
@@ -4,9 +4,15 @@ This is the entry point for the build backend. Items in this module are designed
 
 from __future__ import annotations
 
+import contextlib
 import sys
+from typing import TYPE_CHECKING
 
 from .._compat import tomllib
+
+if TYPE_CHECKING:
+    from collections.abc import Iterator
+    from typing import Literal
 
 __all__ = [
     "build_editable",
@@ -15,9 +21,23 @@ __all__ = [
     "get_requires_for_build_editable",
     "get_requires_for_build_sdist",
     "get_requires_for_build_wheel",
-    "prepare_metadata_for_build_editable",
-    "prepare_metadata_for_build_wheel",
 ]
+
+
+@contextlib.contextmanager
+def _exit_on_failed_live_process() -> Iterator[None]:
+    """Translate a FailedLiveProcessError into a clean ``SystemExit(1)``."""
+    from .._logging import rich_print
+    from ..errors import FailedLiveProcessError
+
+    try:
+        yield
+    except FailedLiveProcessError as err:
+        sys.stdout.flush()
+        rich_print("\n{bold}***", *err.args, color="red", file=sys.stderr)
+        if err.msg:
+            rich_print(err.msg)
+        raise SystemExit(1) from None
 
 
 def build_wheel(
@@ -25,23 +45,15 @@ def build_wheel(
     config_settings: dict[str, list[str] | str] | None = None,
     metadata_directory: str | None = None,
 ) -> str:
-    from .._logging import rich_print
-    from ..errors import FailedLiveProcessError
     from .wheel import _build_wheel_impl
 
-    try:
+    with _exit_on_failed_live_process():
         return _build_wheel_impl(
             wheel_directory,
             config_settings,
             metadata_directory,
             editable=False,
         ).wheel_filename
-    except FailedLiveProcessError as err:
-        sys.stdout.flush()
-        rich_print("\n{bold}***", *err.args, color="red", file=sys.stderr)
-        if err.msg:
-            rich_print(err.msg)
-        raise SystemExit(1) from None
 
 
 def build_editable(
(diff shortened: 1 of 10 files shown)


henryiii:
Everything looks fine, though I researched the `list()` addition with Fable, as I was worried it was a mistake (hint: it's not):

The `list()` is a worthwhile fix, not a mistake — and your understanding of the #1048 trick is the key to why. The pruning trick is the **in-place mutation** of `dirs` (via `dirs.remove(dname)` at `src/scikit_build_core/build/_file_processor.py:94`), and that is untouched. The `list()` only changes what's *iterated*, which fixes the classic remove-while-iterating bug: removing the current element shifts the list, so the iterator skips the next sibling. With consecutive excluded sibling directories, only every other one was actually pruned, and `os.walk` descended into the rest.

This isn't just wasted traversal — it breaks the exact behavior #1048 was built to fix. I verified by running the same scenario against main's `src` and this branch's `src`: a tree with `include=["README.md"]` and four excluded sibling dirs each containing a README:

- **main**: yields `README.md`, **`excl_a/README.md`, `excl_c/README.md`** — the alternating-skip pattern leaks excluded-directory READMEs, which is precisely what #1048's commit message says it was preventing.
- **branch**: yields only `README.md`. 

(There's also a secondary risk: `os.walk(..., followlinks=True)` descending into unpruned excluded trees can hit symlink loops.)

As for why no existing test failed: the #1048 tests never place two excluded sibling directories at the same level — each test excludes at most one directory per level or uses file-level glob patterns, so the skip never triggered. That gap is covered now: the branch commit `08adc5d` added `test_consecutive_excluded_dirs_not_descended` in `tests/test_file_processor.py`, which creates two adjacent excluded siblings and monkeypatches `os.walk` to assert neither is descended into — it fails without the `list()`. All 35 tests in that file pass on the branch.

PR #1325, opened 2026-06-12
chore(tests): reduce test-suite wall-clock time

### LINKED ISSUE #1324: chore: Reduce test-suite wall-clock time
Built a plan with Claude Fable 5 to reduce the time the test suite takes.

# Reduce test-suite wall-clock time

## Context

The full suite (665 tests) runs on ~22 CI matrix jobs; the slowest (Windows) jobs take 13–17
minutes, dominated by the 83 `integration` + 33 `compile` tests. Real data from the latest CI run
(windows-11-arm job, `--durations=20`): individual integration tests cost 35–68s each there.
Three problems stand out:

1. **Non-orthogonal integration tests.** `tests/test_editable_generated.py` is 48 of the 83
   integration tests (8 functions × editable{not_editable, redirect, inplace-skipped} ×
   isolate{2}). Every instance does a fresh venv + `pip install .` of the *identical*
   `cmake_generated` package (~37s each on Windows); only the cheap post-install assertions
   differ. 32 actual installs where 4 suffice.
2. **C++/pybind11 compiles where bindings aren't under test.** 10 sample packages compile C++ via
   pybind11; the tests assert build-system behavior (wheel/sdist layout, editable wiring,
   install dirs), not binding behavior. pybind11 also adds a pip install to every isolated build
   env. MSVC C++ compiles are the big per-test cost on Windows.
3. **Network + slow clone in `sdist_config`**: FetchContent git-clones pybind11 from GitHub
   (~40–60s per variant, 3 test instances).

User decisions (via Q&A): merge the editable_generated tests into one; convert all pybind11
packages to raw C **except** `simple_pyproject_ext` (canonical example, baked sdist hashes —
untouched); make `sdist_config` fetch a tiny local dummy project (this removes its pybind11 use
entirely, so `simple_pyproject_ext` ends up the only pybind11 consumer — `test-pybind11`
CI group stays for it); use uv for auxiliary venv installs with pip fallback.

Estimated effect on the slowest jobs: roughly halves them (editable_generated alone is ~17 min
of CPU on windows-arm; C conversions cut 10–30s off most remaining compile tests; rebuild-test
trim saves ~60s/job).

### 1. Consolidate `tests/test_editable_generated.py` (48 collected → 6, 32 installs → 4)

Merge the 8 test functions into a single `test_generated_files(editable, isolated, isolate)`:
- One `isolated.install("-v", *editable.flags, ".")`, then all 8 assertion groups in sequence,
  each preceded by a comment naming the scenario it documents (keep the module docstring and
  per-scenario explanatory comments — this file deliberately documents supported generated-file
  patterns).
- Keep the shared parametrization: `editable` = not_editable / redirect / inplace-skipped
  (same skip reason), `isolate` = isolated / not_isolated, the 3.9 `importlib.resources.files`
  skipif, and the `compile`/`configure`/`integration` markers.
- All current assertions carry over verbatim (static data, configure-time data, build-time data,
  ctypes resource, configure-time module, generated-module-generated-data, build-time `_version`
  module + `importlib.metadata` version, nested2 subpackage data).

### 2. Trim redundant parametrization elsewhere

- `tests/test_pyproject_pep518.py::test_pep518_rebuild_build_dir` (slowest test: 68s + 55s on
  windows-arm; each variant builds twice): drop the `build_args` parametrization, keep only
  `("--wheel",)` — the sdist→wheel path is already covered by `test_pep518_wheel`, and the
  rebuild logic under test is identical either way.
- `tests/test_editable.py::test_cython_pxd`: restrict `editable` to `["redirect", "inplace"]`
  (`@pytest.mark.parametrize("editable", ["redirect", "inplace"], indirect=True)`) — the
  non-editable pxd path doesn't involve the editable machinery this test exists for. ~41s/run
  on Windows.

### 3. Convert pybind11/C++ sample packages to raw C

Pattern (modeled on `tests/packages/simplest_c/`): replace
`find_package(pybind11 CONFIG REQUIRED)` + `pybind11_add_module(...)` with
`find_package(Python COMPONENTS Interpreter Development.Module REQUIRED)` +
`python_add_library(<target> MODULE <src.c> WITH_SOABI)`, `LANGUAGES CXX → C`, drop `pybind11`
from `build-system.requires`. `WITH_SOABI` on CMake 3.15 is safe: scikit-build-core injects its
FindPython backport for CMake < 3.26.1 (`src/scikit_build_core/builder/builder.py:208`), used by
all three build paths (PEP 517, setuptools shim, hatch plugin); `simplest_c` already passes the
CMake 3.15.x CI rows with this exact pattern.

C module templates: keep the same Python-visible API the tests assert on —
`add`/`subtract`/`__version__` (via `VERSION_INFO` compile definition) for the `cmake_example`
packages; `add`/`subtract` for `_core` packages.

| Package | Source | Test edits |
|---|---|---|
| `hatchling` | `cpp/example.cpp` → `cpp/example.c` | `test_hatchling.py:61` sdist list `example.cpp`→`.c`; drop `"pybind11"` from install at `:196`; remove the two cygwin-pybind11 xfails (`:118`, `:183`) |
| `simple_setuptools_ext` | `src/main.cpp` → `src/main.c` | drop `"pybind11"` from `usefixtures` in `test_setuptools_pep517.py:132,169,433` and `test_setuptools_pep518.py:24,63,91` |
| `toml_setuptools_ext` | same | `test_setuptools_pep517.py:256` |
| `mixed_setuptools` | same (`_core`, keep the 5 `EXAMPLE_DEFINE` checks) | `test_setuptools_pep517.py:292` |
| `plugin_setuptools_install_dir`, `wrapper_setuptools_install_dir` | same (`_core`) | `test_setuptools_pep517.py:332,407` |
| `wrapper_setuptools_classic_layout` | same (`_core`) | `test_setuptools_pep517.py:366` |
| `simple_pyproject_source_dir` | `src/main.cpp` → `src/main.c` | `test_pyproject_pep517.py:259` |
| `cmake_generated` (no pybind11, just CXX→C) | `pkg.cpp` → `pkg.c` (drop `extern "C"`); `wheel.exclude` `**.cpp`→`**.c` in its pyproject.toml; keep `VERSION 1.2.3` on `project()` (regex metadata provider reads it) | none |
| `simple_pure` (CXX→C) | `simple_pure.cpp` → `.c`, drop `cxx_std_11`; stdout must stay exactly `"0 one 2 three \n"` | comment-only updates |

Do **not** touch: `tests/packages/simple_pyproject_ext/` (baked sdist hashes in
`tests/conftest.py:313`), the static file-API fixtures under `tests/api/simple_pure/.cmake/`
(`test_fileapi.py::test_included_dir` asserts `language == "CXX"` against checked-in JSON),
`src/scikit_build_core/_vendor/`.

`tests/utils/download_wheels.py`: split `EXTRA` so pybind11 moves out of it (hatchling tests
install `*download_wheels.EXTRA`) but **stays in `WHEELS`** — `simple_pyproject_ext` isolated
builds still resolve pybind11 from the wheelhouse via `PIP_NO_INDEX`/`PIP_FIND_LINKS`:

```python
EXTRA = [... cmake, ninja as today ...]
PYBIND11 = ["pybind11"] if importlib.util.find_spec("pybind11") is not None else []
WHEELS = [..., *EXTRA, *PYBIND11]
```

No docs or CI changes needed (verified: no `-k` selection anywhere; docs reference
`docs/examples/`, not test packages; `test-pybind11` dependency group still used).

### 4. `sdist_config`: network-free FetchContent + C

Replaces the GitHub pybind11 clone with a tiny committed dummy project; the module becomes raw C
that `#include "dummy.h"` from the fetched content (proving the fetch happened — the build can't
succeed without it). Still exercises everything this package exists for: `sdist.cmake = true`,
FetchContent populating a gitignored dir bundled into the sdist via `sdist.include`, the
"integrated" (pre-fetched) rebuild path, the three `generate` entries, and `cmake.define`
pass-through.

New files in `tests/packages/sdist_config/`:
- `external/dummy/CMakeLists.txt` — INTERFACE library `dummy_lib` exposing `include/`
- `external/dummy/include/dummy.h` — `#define DUMMY_LIFE 42`
- `external/dummy/tools/dummy_helper.cmake` — keeps three parallel `sdist.include` patterns

`CMakeLists.txt`: `LANGUAGES C`; integrated-vs-fetch branch becomes
`message(STATUS "Using integrated dummy")` / `message(STATUS "Fetching dummy project")` with
`FetchContent_Declare(dummy URL "${CMAKE_CURRENT_SOURCE_DIR}/external/dummy" SOURCE_DIR
"${CMAKE_CURRENT_SOURCE_DIR}/dummy")` (local `URL` = directory copy, no network, no hash
warning; works both from the copied workdir and from inside the extracted sdist).
`main.cpp` → `main.c` keeping `life()` returning `DUMMY_LIFE`. pyproject `sdist.include`:
`pybind11/...` → `dummy/...`; `.gitignore`: `/pybind11` → `/dummy`.

`tests/test_pyproject_pep518.py` edits:
- `:102` `"main.cpp"` → `"main.c"`
- `:108` replace the `>= 10 pybind11 files` count with an explicit subset check for the three
  bundled `sdist_config-0.1.0/dummy/...` paths (don't substring-count "dummy" — the committed
  `external/dummy/` files would always match)
- `:139` `"Cloning into 'pybind11'..."` (stderr) → `"Fetching dummy project"` in `out`
- `:141,143` `"Using integrated pybind11"` → `"Using integrated dummy"` (same
  present-for-sdist_to_wheel / absent-for-wheel_directly logic)
- remove the explicit `@pytest.mark.network` on the two sdist_config tests (`:74`, `:117`
  area) — conftest re-adds `network` for `isolated`-fixture tests, but the explicit marker is
  now wrong; wheel assertions (`life() == 42`, 3 top-level names) unchanged
- assert only our own `message(STATUS)` lines, never FetchContent internals (vary across
  CMake 3.15–4.x)

### 5. uv fast path for auxiliary venv installs (`tests/conftest.py`)

`VEnv` gains an aux-install path using `uv pip install --python <venv-python>` (with
`--no-index --find-links <wheelhouse>` when a wheelhouse is set), falling back to pip when
`shutil.which("uv")` is None. Opt-in only — the install-under-test stays on pip:
- add `VEnv.aux_install(*args)` (or `install(..., fast=True)`)
- switch call sites that install tooling, not the package under test:
  `prepare_no_build_isolation` (`pip>23`, `scikit-build-core`, cmake/ninja),
  `isolated.install("build[virtualenv]")` (test_pyproject_pep518.py, test_hatchling.py),
  `isolated.install("cython")` (test_editable.py), the hatchling tooling installs
  (`test_hatchling.py:80,196`).
- every `pip install .` / `-e .` / built-wheel install keeps pip.

## Implementation order (separate commits; could be 2–3 PRs)

1. test consolidation + parametrization trims (sections 1–2) — pure test-file changes, biggest win
2. mechanical C conversions + `download_wheels.py` split (section 3)
3. `sdist_config` redesign (section 4)
4. uv aux-install fast path (section 5)

## Verification

- `uv run pytest tests/test_editable_generated.py -q` — 6 collected (4 run, 2 skipped), all pass
- `uv run pytest tests/test_setuptools_pep517.py tests/test_setuptools_pep518.py tests/test_hatchling.py tests/test_pyproject_pep517.py tests/test_pyproject_pep518.py tests/test_editable.py tests/test_simple_pure.py -n auto` — converted packages
- `uv run pytest tests/test_pyproject_pep518.py -k sdist_config -n auto` with network disabled
  (e.g. unplug or `https_proxy=http://127.0.0.1:9` for the cmake child) to prove no fetch
- spot-check in a venv **without** pybind11: only `simple_pyproject_ext` tests skip
- full suite: `uv run pytest -n auto`; compare `--durations=20` before/after
- `prek -a --quiet`
- after merge, compare CI wall-clock on the windows jobs (expect ~17 min → roughly half)

henryiii:

Closes #1324.

The slowest CI jobs (Windows, 13–17 min) are dominated by the `integration`/`compile` tests; per `--durations=20` on the windows-11-arm job, individual integration tests cost 35–68 s each there. This PR attacks the four biggest sources of redundant work. Locally the full suite now runs in ~32 s with `-n auto` (615 passed, 7 skipped); the Windows CI jobs should land at roughly half their current wall clock.

### 1. Consolidate `test_editable_generated.py` (32 installs → 4)

Its 8 test functions installed the **identical** `cmake_generated` package (8 × editable × isolate = 32 venv + `pip install .` cycles at ~37 s each on Windows CI) and differed only in cheap post-install assertions. They are now a single `test_generated_files` performing one install per (editable, isolate) combination with all assertion groups inline, each labeled with the scenario it documents. Coverage is unchanged.

Also trimmed:
- `test_pep518_rebuild_build_dir` (the single slowest test: 68 s + 55 s) lost its sdist-vs-wheel parametrization — the rebuild logic under test is identical either way and the sdist→wheel path is covered by `test_pep518_wheel`.
- `test_cython_pxd` dropped the `not_editable` mode; it exists to test editable pxd sharing.

### 2. Convert sample packages from pybind11/C++ to raw C

The tests using these packages assert build-system behavior (wheel/sdist layout, editable wiring, install dirs), not binding behavior, so the C++ compile (slow under MSVC) and the pybind11 install in every isolated build env were pure overhead. Converted to tiny C-API modules using `find_package(Python COMPONENTS Interpreter Development.Module)` + `python_add_library(... WITH_SOABI)` — the same pattern `simplest_c` already exercises on the CMake 3.15.x CI rows: `hatchling`, `simple_setuptools_ext`, `toml_setuptools_ext`, `mixed_setuptools`, `plugin_setuptools_install_dir`, both `wrapper_setuptools_*`, and `simple_pyproject_source_dir`; `cmake_generated` and `simple_pure` (no bindings) just switch `LANGUAGES CXX` → `C`.

`simple_pyproject_ext` **stays on pybind11** (canonical `cmake_example`, baked sdist hashes), so pybind11 coverage and the `test-pybind11` dependency group remain. pybind11 stays in the wheelhouse for it via a new `PYBIND11` list in `download_wheels.py`, but is no longer in `EXTRA`, so the hatchling tests stop installing it.

### 3. Make `sdist_config` network-free

It exists to test `sdist.cmake = true` bundling FetchContent-populated (gitignored) sources into the sdist via `sdist.include`, plus the integrated-rebuild path — what it fetched was incidental, yet it git-cloned pybind11 from GitHub on every run (~40–60 s per variant on Windows, network required). It now FetchContent-copies a tiny committed dummy project from a local `URL` (directory URLs are supported in CMake 3.15), and the module `#include`s a header from the fetched content, so a build cannot succeed unless the fetch actually populated the source tree. Verified: no remote URL remains in the package.

### 4. uv fast path for auxiliary venv installs

New `VEnv.aux_install` installs tooling (`build[virtualenv]`, `cython`, `hatchling`, the no-build-isolation preparation) with `uv pip install --python <venv>` against the wheelhouse when uv is on PATH, falling back to pip otherwise (fallback verified by running with uv stripped from PATH). Installs that are themselves under test (`pip install .` / `-e .` / built wheels) stay on pip.

### Verification

- Full suite: `pytest -n auto` — 615 passed, 7 skipped (was 665 collected; consolidation removed ~43 redundant instances).
- With pybind11 uninstalled, only the `simple_pyproject_ext` tests skip.
- `prek -a --quiet` clean; `check-sdist` covers the new committed files.


henryiii:

The Windows failures were not caused by this PR: GitHub rolled `windows-latest` to the `windows-2025-vs2026` image mid-rollout, and CMake < 4.2 cannot detect VS 2026 (falls back to NMake → `CMAKE_C_COMPILER not set`). The fix is in #1326; its commit is cherry-picked onto this branch so CI here can run green and produce the runtime comparison. Once #1326 merges, a rebase drops the duplicate commit.

henryiii:

CI is green with the #1326 pin stacked. Job-time comparison against main's last run ([27300341579](https://github.com/scikit-build/scikit-build-core/actions/runs/27300341579) vs [27396176765](https://github.com/scikit-build/scikit-build-core/actions/runs/27396176765)):

| Job | main | this PR | Δ |
|---|---|---|---|
| 3.12 • CMake 4.0 • windows-11-arm | 16m45s | 10m28s | −38% |
| 3.12 • CMake 3.26.x • win x86 | 16m15s | 8m29s | −48% |
| pypy-3.9 • CMake 3.21.x • windows-2022 | 15m17s | 12m02s | −21% |
| msys UCRT | 13m52s | 10m40s | −23% |
| 3.13 • CMake 3.26.x • windows | 13m47s | 8m12s | −41% |
| 3.14 • CMake 4.0.x • windows | 13m17s | 7m11s | −46% |
| 3.11 • CMake 3.26.x • windows-2022 | 13m15s | 7m32s | −43% |
| Manylinux free-threaded | 11m38s | 6m16s | −46% |
| Min 3.11 • windows | 10m49s | 6m43s | −38% |
| pypy-3.10 • macos | 7m49s | 3m22s | −57% |
| typical ubuntu rows | 3–4m | ~2m | −40–50% |

- **Whole-matrix wall clock** (slowest job): 16m45s → 12m02s; the slowest non-PyPy Windows job drops from ~16m to ~8.5m.
- **Total compute across all jobs**: 16,680 s → 11,196 s (**−33%**).
- The new `windows-2025-vs2026` + CMake 4.2.x row comes in at 8m13s.

Caveats: the previously-`windows-latest` rows also changed runner image (pin from #1326), and there's normal runner variance — but the same 30–50% cuts show up on untouched platforms (ubuntu, macos, manylinux), so the bulk is attributable to the test changes themselves.

tests/utils/download_wheels.py
```diff
@@ -21,8 +21,9 @@
 if importlib.util.find_spec("ninja") is not None:
     EXTRA.append("ninja")
 
-if importlib.util.find_spec("pybind11") is not None:
-    EXTRA.append("pybind11")
+# Only needed by simple_pyproject_ext; kept in the wheelhouse so its isolated
+# builds can resolve it, but not part of EXTRA (tooling installed into venvs).
+PYBIND11 = ["pybind11"] if importlib.util.find_spec("pybind11") is not None else []
```
LecrisUT, line 26: Don't see why this was needed

tests/utils/download_wheels.py
```diff
@@ -21,8 +21,9 @@
 if importlib.util.find_spec("ninja") is not None:
     EXTRA.append("ninja")
 
-if importlib.util.find_spec("pybind11") is not None:
-    EXTRA.append("pybind11")
+# Only needed by simple_pyproject_ext; kept in the wheelhouse so its isolated
```
LecrisUT, line 24: Why do we explicitly need it there 🤔. At least we should have pybind11 covered in `examples`

tests/conftest.py
```diff
@@ -196,6 +198,20 @@ def install(self, *args: str, isolated: bool = True) -> None:
...
+        Uses uv when available, since it is much faster; installs that are
+        themselves under test must use :meth:`install` (pip) instead.
+        """
+        if UV is None:
+            self.install(*args)
+            return
+        cmd = [UV, "pip", "install", f"--python={self.executable}"]
+        if self.wheelhouse is not None:
+            cmd += ["--no-index", f"--find-links={self.wheelhouse}"]
+        self.run(*cmd, *args)
```
LecrisUT: Why not move this into `install` itself?

Although we need at least one editable mode or build isolation that covers both `pip` and `uv` because of the fake venv isolation in `pip`

henryiii:
Asked Claude to reply. Curious to see how it compares to the analysis it gave me.

tests/utils/download_wheels.py
henryiii: :robot: _AI text below_ :robot:

`EXTRA` plays two roles: it is part of `WHEELS` (the wheelhouse download list), but `test_hatchling.py` also installs `*download_wheels.EXTRA` directly into venvs for its `--no-isolation` builds. After the C conversion the hatchling package no longer needs pybind11 there — but `simple_pyproject_ext`'s isolated builds still resolve `pybind11` from the wheelhouse (`PIP_NO_INDEX` + `PIP_FIND_LINKS`), so it has to stay in `WHEELS` while leaving `EXTRA`.

Folding it back into `EXTRA` would also work; the only cost is installing pybind11 into every hatchling test venv for no reason. Happy to collapse it if you prefer the simpler file.

tests/utils/download_wheels.py
henryiii: :robot: _AI text below_ :robot:

pybind11 keeps real coverage in both places:

- `tests/packages/simple_pyproject_ext` (the canonical pybind11 package, with baked sdist hashes) is deliberately untouched and still runs through the main flows — `test_pep518_sdist`/`_wheel`/`_rebuild_build_dir`/`_pip` plus the PEP 517 tests.
- The "Check examples" CI job (`nox -s test_doc_examples`) does full builds of `docs/examples/getting_started/pybind11` and `docs/examples/downstream/pybind11_example`.

The conversions only touched packages whose tests assert build-system behavior (layout, install dirs, editable wiring), not binding behavior.

tests/conftest.py
henryiii: :robot: _AI text below_ :robot:

It's a deliberate split: `install` is the system under test in most of these tests — pip's PEP 517 frontend behavior (its sys.path-based "fake venv" build isolation, its editable wheel handling) is exactly what's being exercised. If `install` silently switched to uv when present, what the suite tests would depend on what's on PATH, and uv's real-venv isolation behaves differently. So: `aux_install` = tooling setup, fastest path wins; `install` = deterministic frontend under test.

Agreed on the second point — testing uv as a frontend in at least one editable and one build-isolation case is worth doing, but it's a coverage expansion rather than part of this speed-focused PR. Opened #1328 to track it.

tests/utils/download_wheels.py
LecrisUT: > Folding it back into `EXTRA` would also work; the only cost is installing pybind11 into every hatchling test venv for no reason.

It didn't do well

tests/utils/download_wheels.py
LecrisUT: > The conversions only touched packages whose tests assert build-system behavior (layout, install dirs, editable wiring), not binding behavior.

Yeah it didn't answer the question. Do we need it to be covered in the pytest test suite in the first place? Would be 1 less circular dependency to worry in there

tests/utils/download_wheels.py
henryiii: ~~What do you not see is needed? We only download a pybind11 wheel if we already have pybind1 (that's not new, was added some time ago). This is...~~ Ahh, you mean not using EXTRA but renaming PYBIND11? I didn't understand what you meant at first.

We use `EXTRA` in `tests/test_hatchling.py:80`. If this was left in it, then hatchling's test would require pybind11 for no reason.

```console
$ git grep EXTRA
tests/test_hatchling.py:80:        "build[virtualenv]", "scikit-build-core", "hatchling", *download_wheels.EXTRA
tests/test_hatchling.py:184:        *download_wheels.EXTRA,
```

tests/utils/download_wheels.py
henryiii: I'd like to keep pybind11 as something we test, it's important to me. :) It's also testing building from a downloaded package, which the raw C examples don't do. But C++ is much slower than C to compile, so we only need 1-2 tests with pybind11. I like writing pybind11 better than raw CPython, which is why we used it so heavily before.

tests/utils/download_wheels.py
LecrisUT: Aha. That explains it simpler :slightly_smiling_face:. I guess remaining part is https://github.com/scikit-build/scikit-build-core/pull/1325#discussion_r3401441480

tests/conftest.py
henryiii: Described above, but using uv for package installs is faster, and we are keeping pip for the core install. If we moved to uv there too, then we wouldn't be exposed to pip's internal hacks. Those _might_ get removed, someone (Richard Si if I remember correctly?) has some funding to work on pip and this is the top priority.

tests/utils/download_wheels.py
LecrisUT: Sure, but don't we have the coverage in the examples already? Is it needed for the test matrix in the CI?

Main thing is if this should not be its separate standalone test.

tests/conftest.py
LecrisUT: Not sure if it deserves its dedicated `aux_install` instead of being a switch inside `Venv` or the fixtures. I am considering more that this should be the default unless we really test a pip thing.

> If install silently switched to uv when present, what the suite tests would depend on what's on PATH, and uv's real-venv isolation behaves differently.

This part does not seem right. It should be installing in the same venv in both approaches. That is to say, I don't see the part that is blocking for this part to be in the `install` command.

tests/utils/download_wheels.py
henryiii: I think having an extra package (other than scikit-build-core) and using something (ideally header files or CMake files) from that package being used is useful. It also uses our lookup system for config modules.

LecrisUT: Hmm, fair enough

henryiii: (I'm open to others, but don't know of one. I know it does have the circular issue, but that's not new here ;) )

tests/conftest.py
henryiii: :robot: _AI text below_ :robot:

Reworked in 5f0a106: `aux_install` is gone, replaced by an `installer=` argument on `install` itself. The default (`"auto"`) uses uv when available; the installs that are themselves under test (`.`/`-e .`) pin `installer="pip"` explicitly at the call site. An explicit `installer="uv"` is also accepted, which gives #1328 a natural hook for parametrizing the frontend later.

henryiii: Running `pip install .` is the thing you don't want to move to uv. Pip has bugs that uv doesn't related to build isolation. But everything else you can use uv or pip, it doesn't matter and uv is faster.

I've combined this to one function, with an `installer=` arg.

LecrisUT (approved):

tests/conftest.py
henryiii: Hey, I didn't ask for that second comment. Once you tell it it's okay to comment, the floodgates are open, I think. :) I like my little AI text header so no one thinks AI is my writing. :)

henryiii: Fun fact: A little more installs with uv now. :)

PR #1337, opened 2026-06-18
fix: don't treat versioned sonames as importable modules in editable installs (#1144)

### LINKED ISSUE #1144: Invalid soname for known_wheel_files in editable mode
We noticed an issue on one of our CI runner when building our project in editable mode.
Importing the module fails.
I tracked down the issue to the `known_wheel_files` dict passed to `install` in the `_pytango_editable.py` file imported in `_pytango_editable.pth`: ` {'tango._tango': 'tango/_tango.so.10.1.0.0'}` (see [log](https://gitlab.com/tango-controls/pytango/-/jobs/11321752584))

```
$ python -c 'import sys;print(sys.path)'
['', '/opt/conda/lib/python313.zip', '/opt/conda/lib/python3.13', '/opt/conda/lib/python3.13/lib-dynload', '/opt/conda/lib/python3.13/site-packages', '/builds/tango-controls/pytango']
$ python -c 'import tango;print(tango.utils.info())'
  File "<string>", line 1, in <module>
    import tango;print(tango.utils.info())
    ^^^^^^^^^^^^
  File "/builds/tango-controls/pytango/tango/__init__.py", line 247, in <module>
    from ._tango import (
    ...<179 lines>...
    )
ModuleNotFoundError: No module named 'tango._tango'
```

When working, we have: `{'tango._tango': 'tango/_tango.so'}`([log](https://gitlab.com/tango-controls/pytango/-/jobs/11321752592))

In both cases, we have the same files under `site-packages/tango`:

```
$ ls -la /opt/conda/lib/python3.13/site-packages/tango
total 77804
drwxr-xr-x   2 root root     4096 Sep 11 13:09 .
drwxr-xr-x 127 root root     4096 Sep 11 13:09 ..
-rw-r--r--   1 root root 26550824 Sep 11 13:09 _tango.so
-rw-r--r--   1 root root 26550824 Sep 11 13:09 _tango.so.10
-rw-r--r--   1 root root 26550824 Sep 11 13:09 _tango.so.10.1.0.0
```

My understanding is that `_tango.so.10.1.0.0` isn't a valid [PEP 3149-style filename](https://peps.python.org/pep-3149/) as it doesn't end with `.so`, meaning that `_tango` isn't recognised as a loadable extension.

`scikit-build-core` should always use the `.so` file.

henryiii:

## Summary

In editable installs, the redirect maps (`known_source_files` / `known_wheel_files`) serve **two** roles:

1. **Module resolution** — `find_spec` looks a module name up in the maps to find the file to load.
2. **Resource directories** — every tracked file contributes its parent directory to `submodule_search_locations`, so `importlib.resources` can find data files.

`is_valid_module` was overloaded for both roles. Because `path_to_module` keys on the stem before the first `.`, the SONAME variants Linux CMake builds commonly produce side-by-side — `_tango.so`, `_tango.so.10`, `_tango.so.10.1.0.0` — all mapped to the same key `tango._tango`, and `libdir_to_installed` had no collision handling, so a non-importable versioned soname could win non-deterministically. The result was `ModuleNotFoundError: No module named 'tango._tango'` (issue #1144).

## Fix

Split the two roles instead of overloading one predicate:

- **`is_valid_module`** stays broad: it tracks data/resource files (`.txt`, `.pyx`, `.pxd`, ...) **and** versioned sonames, so their directories are still registered and `importlib.resources` keeps working.
- **`module_loader_rank`** (new) ranks a file by Python's own import-loader precedence, using `importlib.machinery` suffixes: extension module (`.so`/`.pyd`/`.abi3.so`/...) → source (`.py`) → bytecode (`.pyc`) → non-importable (data files and versioned sonames, which rank last). This mirrors the order CPython's `FileFinder` tries loaders.
- **`_prefer_module`** is a collision tiebreaker, applied in **both** `mapping_to_modules` and `libdir_to_installed`: for a given module name it keeps the highest-precedence file. So editable installs resolve the **same file a real wheel would import** — the plain `.so` beats versioned sonames (issue #1144), and an extension module beats a `.py` sitting beside it (rather than the reverse).

Earlier iterations used a simpler boolean "is this importable?" check that preferred `.py`/`.pyc`; that diverged from CPython's loader order when an extension module and a source file shared a name (raised in review). Ranking by loader precedence makes editable and wheel imports agree.

`module_loader_rank` relies on `importlib.machinery.EXTENSION_SUFFIXES`, which already contains the native extension on each platform (`.so` on Unix, `.pyd` on Windows) — matching the fact that an editable install resolves on the platform it was built for.

Fixes #1144

## Testing

Unit tests in `tests/test_editable_unit.py` (no virtualenv required):

- `test_is_valid_module` — modules, data files, and versioned sonames are all tracked; non-identifier paths are rejected.
- `test_is_module` / `test_is_module_rejects_versioned_sonames` — only real module extensions are importable; sonames, data files, and Cython sources are not. (The soname test is skipped where `.so` is not importable, i.e. Windows, since versioned sonames are a Unix concept.)
- `test_module_loader_rank_matches_import_precedence` — extension < source < bytecode < non-importable.
- `test_libdir_to_installed_prefers_extension_over_source` — a dir with both `mod.<ext>` and `mod.py` resolves `pkg.mod` to the extension module, matching wheel import order.
- `test_libdir_to_installed_prefers_plain_so_over_versioned` — a `tango/` dir with all three soname variants resolves `tango._tango` to `tango/_tango.so`.

The existing `test_navigate_editable_pkg` (with virtualenv) continues to pass, confirming data/resource files remain resolvable via `importlib.resources`.

---

## Follow-up: separate directory tracking from module resolution

The redirect maps were doing double duty — resolving importable modules **and**, as a side effect, registering each file's directory so `importlib.resources` can navigate it. To get the directory side, non-importable data files were stored as fake module entries (`pkg.resources.file → file.txt`); `find_spec` could then hand one of those back as a module's origin (e.g. `import pkg.foo` returning `foo.txt`).

These are now split:

- `mapping_to_modules` / `libdir_to_installed` return **importable files only** (used by `find_spec`).
- A new `collect_search_locations` walks every tracked file — modules and data alike — to build the package→`__path__` map (`known_directories`) and the set of regular packages (`known_packages`, including `.pxd`/`.pyx` `__init__` packages). These are passed to the finder, which no longer derives `__path__` from the file maps.

Net effect: a non-importable file is never resolved as a module, while data-only directories (e.g. a package dir where CMake installs only data, with no module of its own in that tree) still join their package's `__path__` so `importlib.resources` can reach them — verified by `test_navigate_editable_pkg` and a new `test_collect_search_locations_data_only_install_dir`.

This changes the internal `install(...)` signature in the generated `_editable_skbc_*.py` shim (adds `known_directories` / `known_packages`); the shim and its `install(...)` call are emitted together, so there is no version skew.

### DIFF
--- a/src/scikit_build_core/build/_editable.py
+++ b/src/scikit_build_core/build/_editable.py
@@ -8,6 +8,7 @@ from pathlib import Path
 
 from ..resources import resources
 from ._pathutil import (
+    is_module,
     is_valid_module,
     module_loader_rank,
     path_to_module,
@@ -20,6 +21,7 @@ if typing.TYPE_CHECKING:
     from ..settings.skbuild_model import ScikitBuildSettings
 
 __all__ = [
+    "collect_search_locations",
     "editable_inplace_files",
     "editable_redirect",
     "editable_redirect_files",
@@ -37,6 +39,8 @@ def editable_redirect(
     *,
     modules: dict[str, str],
     installed: dict[str, str],
+    directories: dict[str, list[str]],
+    packages: Sequence[str],
     reload_dir: Path | None,
     rebuild: bool,
     verbose: bool,
@@ -59,6 +63,8 @@ def editable_redirect(
     arguments = (
         modules,
         installed,
+        directories,
+        list(packages),
         os.fspath(reload_dir) if reload_dir else None,
         rebuild,
         verbose,
@@ -99,12 +105,15 @@ def editable_redirect_files(
         use_start = sys.version_info >= (3, 15)
     modules = mapping_to_modules(mapping, libdir)
     installed = libdir_to_installed(libdir)
+    directories, known_packages = collect_search_locations(mapping, libdir)
     if settings.editable.rebuild and settings.wheel.install_dir.startswith("/"):
         msg = "Editable installs cannot rebuild an absolute wheel.install-dir. Use an override to change if needed."
         raise AssertionError(msg)
     editable_txt = editable_redirect(
         modules=modules,
         installed=installed,
+        directories=directories,
+        packages=known_packages,
         reload_dir=reload_dir,
         rebuild=settings.editable.rebuild,
         verbose=settings.editable.verbose,
@@ -161,13 +170,17 @@ def get_packages(
 
 def mapping_to_modules(mapping: dict[str, str], libdir: Path) -> dict[str, str]:
     """
-    Convert a mapping of files to modules to a mapping of modules to installed files.
+    Map importable module names to their (absolute) source files.
+
+    Only importable files are included; data/resource files are tracked
+    separately by :func:`collect_search_locations` so that ``find_spec`` never
+    resolves a name to a non-importable file.
     """
     result: dict[str, str] = {}
     selected: dict[str, Path] = {}
     for k, v in mapping.items():
         rel = Path(v).relative_to(libdir)
-        if not is_valid_module(rel):
+        if not is_valid_module(rel) or not is_module(rel):
             continue
         module = path_to_module(rel)
         if module in result and not _prefer_module(rel, selected[module]):
@@ -180,13 +193,16 @@ def mapping_to_modules(mapping: dict[str, str], libdir: Path) -> dict[str, str]:
 
 def libdir_to_installed(libdir: Path) -> dict[str, str]:
     """
-    Convert a mapping of files to modules to a mapping of modules to installed files.
+    Map importable module names to their installed files (relative to ``libdir``).
+
+    Only importable files are included; data/resource files are tracked
+    separately by :func:`collect_search_locations`.
     """
     result: dict[str, str] = {}
     selected: dict[str, Path] = {}
     for v in scantree(libdir):
         pth = v.relative_to(libdir)
-        if not is_valid_module(pth):
+        if not is_valid_module(pth) or not is_module(pth):
             continue
         module = path_to_module(pth)
         if module in result and not _prefer_module(pth, selected[module]):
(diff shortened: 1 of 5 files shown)


henryiii:
Not sure why this fix doesn't match the text above... Edit: Ah, it's in the commit message. I'll investigate.

henryiii:
Codex GPT 5.5 review:

The patch fixes the versioned-soname case but introduces a deterministic preference that can make editable installs import a different file than a normal wheel when multiple importable suffixes share a module name.

Review comment:

- [P2] Respect importlib's module suffix precedence — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/_editable.py:210-210
  When the same installed module stem has multiple importable files, this tiebreaker does not match Python's loader order: `.py`/`.pyc` can replace an extension module (and `.pyc` can replace `.py`) just because this returns true for those suffixes. For an editable install whose CMake install tree contains both `pkg/mod.so` and `pkg/mod.py`, `libdir_to_installed()` records the `.py`, while a normal wheel import would load the extension first, so editable behavior can differ from the built wheel.

henryiii:
GPT 5.5 again:

The patch fixes the main versioned-soname case, but still does not fully match Python import precedence when multiple supported extension suffixes exist for the same module.

Review comment:

- [P3] Rank extension suffixes individually — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/_pathutil.py:120-122
  When a package directory contains more than one valid extension-module filename for the same module, such as `mod.abi3.so` and `mod.cpython-313-...so`, Python's `FileFinder` tries `EXTENSION_SUFFIXES` in order, but this assigns all extension suffixes the same rank and keeps whichever file was scanned first. That can make editable installs resolve a different extension than a normal wheel import would.

henryiii:

Good catch — fixed in a86f315.

`module_loader_rank` was bucketing all extension suffixes into a single rank (0), so when a directory held more than one extension file for the same module (`mod.cpython-313-…so` and `mod.abi3.so`), the winner was whichever `scantree` reached first.

It now ranks by a file's **exact** suffix (everything after the first `.`) and its **index** in the interpreter's ordered suffix list — `EXTENSION_SUFFIXES` (most-specific tag first), then `SOURCE_SUFFIXES`, then `BYTECODE_SUFFIXES` — i.e. the order `FileFinder` itself tries them. So `mod.cpython-313-…so` ⟶ `mod.abi3.so` ⟶ `mod.so` ⟶ `mod.py` ⟶ `mod.pyc`, matching a real wheel import.

A side benefit: `is_module` is now an exact-suffix test instead of `endswith`, so a foreign-ABI file like `mod.cpython-313-…so` on a 3.14 interpreter is correctly treated as non-importable (FileFinder would never load it as `mod` there either).

New tests: `test_module_loader_rank_orders_extension_suffixes` and `test_libdir_to_installed_prefers_specific_extension_tag` (both skip where there are fewer than two extension suffixes, e.g. some platforms).

PR #1347, opened 2026-06-22
docs: show config-settings and env var forms in config reference

### LINKED ISSUE #1147: Improve usability of config reference section
Following discussion with @LecrisUT , we suggest to improve the [config reference](https://scikit-build-core.readthedocs.io/en/latest/reference/configs.html) section to:
* describe how to use each in the context of `pyproject.toml`, `config-settings` and environment (where applicable)
* include a link back to the main documentation if relevant

henryiii:

Closes #1147 (first item: showing how each option is used across `pyproject.toml`, `config-settings`, and environment).

## What this does

Each entry in the [config reference](https://scikit-build-core.readthedocs.io/en/latest/reference/configs.html) now shows all three usage forms as fields, e.g.:

```rst
.. confval:: build-dir
  :type: ``str``
  :config-settings: ``build-dir`` or ``skbuild.build-dir``
  :env: ``SKBUILD_BUILD_DIR``
```

- **`:config-settings:`** lists both the bare key and the namespaced `skbuild.` form, since both are accepted.
- **`:env:`** shows the derived `SKBUILD_*` variable.
- Both flat forms are skipped where they cannot be expressed: arrays of tables (`generate[]`) and nested mappings (`metadata`) — those are `pyproject.toml`-only.

Sphinx's built-in `confval` directive only renders `:type:`/`:default:`, so a small `confval_env` extension (in `docs/ext/`) overrides it to accept and render the two extra fields in the same field list.

## Docstring fix

While adding `:env:`, I found the override-only fields (`fail`, `wheel.tags`, `cmake.toolchain-file`, the `variant*` fields) claimed they were *"only allowed in overrides or config-settings"* — but environment variables work too (`_validate_overrides` treats env as a dynamic source). Reworded to the actual restriction: they cannot be hardcoded in the static `[tool.scikit-build]` table.

## Not included

The other two asks in #1147 (schema crosslinks, backlinks to the main docs) are left for a follow-up.

## Verification

- `nox -t gen` regenerates `README.md`, `docs/reference/configs.md`, and `scikit-build.schema.json` cleanly.
- `nox -s docs` builds with no warnings; the new fields render correctly.
- `tests/test_settings_docs.py` updated; full settings test suite passes.
- `prek -a` clean.


src/scikit_build_core/settings/skbuild_docs_sphinx.py
```diff
@@ -55,7 +80,7 @@ class Item:
     TEMPLATE: typing.ClassVar[str] = textwrap.dedent("""\
     ```{{eval-rst}}
     .. confval:: {item.name}
-      :type: ``{item.type}``{sphinx_default}
+      :type: ``{item.type}``{sphinx_default}{sphinx_config_settings}{sphinx_env}
```
LecrisUT: Let's at least expand the `sphinx_config_settings`

docs/ext/confval_env.py
```diff
@@ -0,0 +1,68 @@
...
+                continue
+            parsed, msgs = self.parse_inline(self.options[key], lineno=self.lineno)
+            field_list.append(
+                nodes.field(
+                    "",
+                    nodes.field_name("", label),
+                    nodes.field_body("", *parsed),
+                )
+            )
+            field_list += msgs
```
LecrisUT: We have `conftabs`. The approach I was thinking is to use that or equivalent. Handling bare `nodes` is very magic for a sphinx extension, would prefer to avoid it like the plague.

LecrisUT (commented):
My thought of how to approach this was to make a proper domain out of it and have a `autodoc` "extension" to it or at least mimic the `conftabs` construction/rendering method.

henryiii:

Thanks for the review. Dropped the `confval_env` extension entirely (fe5fb43) — no more hand-built docutils nodes / directive override.

Instead, the generator now emits every field (Type, Default, Config-settings, Environment variable) as a plain rST field list inside the `confval` body, produced as generated text the same way `conftabs` builds its block. The rendered output is unchanged — a single field list per option — and the `confval-<name>` anchors and `:confval:` cross-references still work. The cramped `{sphinx_default}{sphinx_config_settings}{sphinx_env}` template line is gone; the field list is built explicitly in `Item.fields()`.

On the larger idea (a proper domain + `autodoc`-style extension): agreed that would be the clean long-term home for this, but it's a bigger refactor than this docs PR — happy to track it as a follow-up if you want.

henryiii:
Not sure how to control the AI reply. Especially stuff like "agreed that would be the clean long-term home for this, but". Glad I have the AI disclaimer on any text it writes. :)

LecrisUT:
> Not sure how to control the AI reply. Especially stuff like "agreed that would be the clean long-term home for this, but". Glad I have the AI disclaimer on any text it writes. :)

And I would even put "clean" in quotations. That thing is a doozy to write even the bare minimum. The only advantage it has is to move the whole `skbuild_docs_sphinx.py` to a more sensible place one would expect it in. It is quite a fun rabbit hole to design and dig into though and I've done it at least once [^1].

[^1]: https://github.com/teemtee/tmt/tree/main/docs/ext/tmt_domain

docs/ext/confval_env.py
henryiii: I think having a tab for every one of these would be pretty heavy for a reference. The current form renders pretty well as reference, and I think tabs make more sense on the user-facing page.

LecrisUT: Doesn't look very navigable to me

image

we have the example at the top, but the user would have to scroll up and down to see the example usage, particularly for the config-settings.

And for more complex items like [`cmake.define`](https://scikit-build-core--1347.org.readthedocs.build/en/1347/reference/configs.html#confval-cmake.define) it is more valuable to see the proper example rendered in that reference

henryiii: The examples are in https://scikit-build-core.readthedocs.io/en/latest/configuration/index.html, this is only a reference page.

henryiii:
Let's put this in as a net improvement, than we can followup with something more if that's better. Though I think improving the user-facing page for config is higher priority.

PR #1364, opened 2026-06-24
feat: add force-include for sdist, wheel, and build targets

henryiii:

## Summary

Adds a top-level `[tool.scikit-build.force-include]` table that maps source paths to destinations, modeled on hatchling's `force-include` but extended for scikit-build-core's multi-tree wheel layout and CMake build directory.

```toml
[tool.scikit-build.force-include]
"vendor/lib.so"   = "pkg/_lib.so"                          # bare string -> wheel (platlib)
"tools/run.sh"    = {wheel = "/scripts/run.sh"}            # leading-slash -> wheel tree (experimental)
"../shared/data"  = {sdist = "data", wheel = "pkg/data"}   # per-target; dirs recurse
"toolchain.cmake" = {build = "toolchain.cmake"}            # into CMake build dir, before configure
```

### Semantics
- **Key = source**, relative to the project root; may point outside it (`../shared`), `~` is expanded. File or directory (directories recurse, skipping VCS/`__pycache__` junk via the existing `EXCLUDE_LINES`).
- **Bare string value = wheel destination** (the common case), relative to platlib. A leading `/data`, `/scripts`, `/headers`, `/platlib`, or `/metadata` targets that wheel tree instead (requires `experimental = true`, like `wheel.install-dir`).
- **Inline table** gives per-target control over any subset of `sdist`, `wheel`, and `build`. This form is TOML-only (force-include keys are file paths with dots/slashes that don't round-trip through config-settings dotted keys); the bare-string/wheel form also works via config-settings and `SKBUILD_FORCE_INCLUDE`.
- Force-included files are copied **last**, so they override discovered package files and CMake output at the same destination.
- **A missing source is silently skipped, not an error** (matches hatchling): when building a wheel from an SDist an external source may be gone but is assumed already present at the destination. Use the `sdist` target to vendor it into the tarball so it is.

### Implementation notes
- Taught the settings machinery to route a TOML table to a dataclass member of a union (`Dict[str, Union[str, ForceIncludeTargets]]`) — previously `TOMLSource.convert()` rejected a `dict` value in a union. `json_schema.py` already produced the right `oneOf`; `documentation.py` learned to render a non-optional `Union`, and a one-line guard in `skbuild_schema.py` tolerates the `patternProperties` shape (like `metadata`).
- The leading-slash wheel-tree resolution is factored into `resolve_wheel_tree` and reused by `wheel.install-dir`, the force-include wheel path, and the Hatch plugin (removed a triplicated block).
- Non-wheel destinations are validated to reject absolute or `..`-containing paths so `sdist`/`build` targets can't escape their root.
- The experimental **Hatch** plugin rejects `force-include` in `_validate` (it cannot honor the `sdist` target, which Hatch builds separately), consistent with how `generate`/`metadata` are rejected.

### Tests / docs
- New `tests/test_force_include.py` covers file / directory-recursion / external (`../`) / leading-slash wheel tree / override / missing-source / sdist / build-into-CMake, escape-rejection, plus settings parsing across TOML / env / config-settings.
- A Hatch `_validate` rejection test.
- Generated schema + `docs/reference/configs.md` (`nox -t gen`) and a narrative "Force-including files" docs section.

### Open question
The inline-table form is intentionally TOML-only (see above). Flagging in case you'd prefer a config-settings encoding too.

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -378,22 +378,19 @@ file or a directory, and directories are copied recursively (skipping VCS and
 the platlib (the package area). Force-included files are placed last, so they
 override discovered package files and CMake output at the same destination.
 
-For more control, the value can be an inline table targeting any subset of the
-`sdist`, `wheel`, and `build` outputs independently:
+For more control, the value can be an inline table targeting the `sdist` and
+`wheel` outputs independently:
 
 ```toml
 [tool.scikit-build.force-include]
 # vendor an external data file into both the SDist and the wheel
 "../shared/data.json" = {sdist = "mypackage/data.json", wheel = "mypackage/data.json"}
-# drop a file into the CMake build directory before configure, for CMake to use
-"toolchain.cmake" = {build = "toolchain.cmake"}
 ```
 
 The `wheel` destination also accepts a leading `/data`, `/scripts`, `/headers`,
 or `/metadata` to target that wheel tree instead of the platlib (this requires
 `experimental = true`, like `wheel.install-dir`). The `sdist` destination is
-relative to the SDist root, and `build` is relative to the CMake build
-directory.
+relative to the SDist root.
 
 A missing source is an error by default, unless the entry sets
 `missing-ok = true` (the bare-string form always does, like hatchling):
@@ -403,11 +400,11 @@ A missing source is an error by default, unless the entry sets
 "../maybe-built.so" = {wheel = "mypackage/_lib.so", missing-ok = true}
 ```
 
-When a wheel or build is produced from an SDist rather than a source tree, an
-entry with an `sdist` destination is read from that vendored SDist location
-instead of the original source. So vendoring an external source into the SDist
-(via the `sdist` target) keeps the wheel buildable from the SDist, where the
-original `../` source is gone.
+When a wheel is built from an SDist rather than a source tree, an entry with an
+`sdist` destination is read from that vendored SDist location instead of the
+original source. So vendoring an external source into the SDist (via the `sdist`
+target) keeps the wheel buildable from the SDist, where the original `../`
+source is gone.
 
 The inline-table form is only available in `pyproject.toml`; the bare-string
 (wheel) form can also be set via config-settings or `SKBUILD_FORCE_INCLUDE`.
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -394,10 +394,6 @@ destination. A force-included _directory_ stays subject to that exclude, so a
 bulk tree copy can still be trimmed by an exclude pattern (e.g. force-include a
 directory and exclude `**/*.bzl` to drop the Bazel files from it).
 
-Both tables can be set through any config source, including config-settings and
-the `SKBUILD_SDIST_FORCE_INCLUDE` / `SKBUILD_WHEEL_FORCE_INCLUDE` environment
-variables.
-
 #### Building a wheel from an SDist
 
 A common pattern vendors an external (`../`) source into the SDist and then
@@ -421,9 +417,9 @@ on-disk file always wins, so the vendored copy is preferred when present.
 
 For cases the automatic resolution cannot express — e.g. the wheel source is the
 _original_ external path rather than the SDist output — use
-[overrides](#overrides) keyed on `from-sdist`. Because override dicts merge
-rather than replace, keep the external entry in the `from-sdist = false`
-override (not the base table):
+[overrides](#overrides) keyed on `from-sdist`, with a separate
+`wheel.force-include` entry gated on each build mode (source tree vs.
+wheel-from-SDist):
 
 ```toml
 [tool.scikit-build.sdist.force-include]
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -359,6 +359,49 @@ assume `wheel.platlib = false` (purelib targeted instead).
 
 :::
 
+### Force-including files
+
+Sometimes you need to place a specific file (or directory) at a specific path in
+a distribution, even if it lives outside your package tree or is produced
+elsewhere. `force-include` is a table mapping source paths to destinations:
+
+```toml
+[tool.scikit-build.force-include]
+"vendor/lib.so" = "mypackage/_lib.so"
+"assets" = "mypackage/assets"
+```
+
+The keys are source paths relative to the project root; they may point outside
+it (e.g. `../shared`), and `~` is expanded. A source may be a file or a
+directory, and directories are copied recursively (skipping VCS and
+`__pycache__` junk). A bare string value is the wheel destination, relative to
+the platlib (the package area). Force-included files are placed last, so they
+override discovered package files and CMake output at the same destination.
+
+For more control, the value can be an inline table targeting any subset of the
+`sdist`, `wheel`, and `build` outputs independently:
+
+```toml
+[tool.scikit-build.force-include]
+# vendor an external data file into both the SDist and the wheel
+"../shared/data.json" = {sdist = "mypackage/data.json", wheel = "mypackage/data.json"}
+# drop a file into the CMake build directory before configure, for CMake to use
+"toolchain.cmake" = {build = "toolchain.cmake"}
+```
+
+The `wheel` destination also accepts a leading `/data`, `/scripts`, `/headers`,
+or `/metadata` to target that wheel tree instead of the platlib (this requires
+`experimental = true`, like `wheel.install-dir`). The `sdist` destination is
+relative to the SDist root, and `build` is relative to the CMake build
+directory.
+
+A missing source is silently skipped, rather than erroring: when a wheel is
+built from an SDist an external source may be gone, but it is assumed to already
+be present at the destination (use the `sdist` target to vendor it into the
+SDist so it is). The inline-table form is only available in `pyproject.toml`;
+the bare-string (wheel) form can also be set via config-settings or
+`SKBUILD_FORCE_INCLUDE`.
+
 ## Customizing the output wheel
 
 The python API tags for your wheel will be correct assuming you are building a
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -363,53 +363,55 @@ assume `wheel.platlib = false` (purelib targeted instead).
 
 Sometimes you need to place a specific file (or directory) at a specific path in
 a distribution, even if it lives outside your package tree or is produced
-elsewhere. `force-include` is a table mapping source paths to destinations:
+elsewhere. Each distribution has its own `force-include` table mapping source
+paths to destinations:
 
 ```toml
-[tool.scikit-build.force-include]
-"vendor/extra.cmake" = "cmake/extra.cmake"
-"assets" = "share/assets"
+[tool.scikit-build.sdist.force-include]
+"../shared/data.json" = "mypackage/data.json"
+
+[tool.scikit-build.wheel.force-include]
+"vendor/lib.so" = "mypackage/_lib.so"
+"tools/run.sh"  = "/scripts/run.sh"
 ```
 
 The keys are source paths relative to the project root; they may point outside
 it (e.g. `../shared`) or be absolute, and `~` is expanded. A source may be a
 file or a directory, and directories are copied recursively (skipping VCS and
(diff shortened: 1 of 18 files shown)


henryiii:
GPT 5.5 review in codex:

The new force-include feature misses important frontend and wheel-script scenarios, and its destination validation can escape the intended base on Windows. These are functional issues in the new behavior.

Full review comments:

- [P2] Include metadata force-includes in prepare_metadata — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/wheel.py:500-505
  When an entry targets `wheel = "/metadata/..."` and a frontend calls `prepare_metadata_for_build_wheel`/`prepare_metadata_for_build_editable` first, the metadata-only path returns before this force-include block runs, so the prepared `.dist-info` omits these files. The final wheel then has different metadata contents (or the later metadata comparison raises on the missing file), which violates the PEP 517 metadata contract for frontends that pass `metadata_directory`.

- [P2] Run script processing after force-includes — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/wheel.py:506-510
  When `force-include` targets `/scripts/...` with a Python shebang, the file is copied here after `process_script_dir(wheel_dirs["scripts"])` has already run, so its shebang is not normalized to `#!python`. Wheel installers rely on that normalization for scripts, so force-included scripts should be copied before processing or the script directory should be processed again afterward.

- [P2] Reject Windows absolute and parent destinations — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/_pathutil.py:156-165
  On Windows, destinations like `C:/tmp/x` or `..\x` are not rejected by `PurePosixPath`, but `Path(base) / dest` treats the drive/backslash as filesystem syntax and can copy outside the intended wheel/build/sdist base. Since the setting is documented as requiring a relative destination without parent components, this validation should also reject Windows drives/backslashes (or normalize before joining).

henryiii:

Addressed all three P2 findings in `ebfdb4d`:

- **Metadata force-includes in prepare_metadata** — extracted `_force_include_into_wheel()` and now run it with `only_metadata=True` in the metadata-only path, before `dist_info_contents()`. The prepared `.dist-info` now matches the final wheel, so the metadata comparison no longer raises for frontends passing `metadata_directory`.
- **Script processing after force-includes** — moved `process_script_dir()` to run *after* the wheel force-include copy, so a force-included `/scripts/...` file with a python shebang gets normalized to `#!python`.
- **Windows-escaping destinations** — `iter_force_include` now also rejects a backslash and a Windows drive (e.g. `C:/x`), which `PurePosixPath` missed but `base / dest` would honor on Windows.

New tests cover each: `test_force_include_metadata_in_prepare`, `test_force_include_script_shebang_normalized`, and `C:/escape.txt`/`..\\escape.txt` cases added to the escape-rejection parametrization.

henryiii:
> pytorch is moving to scikit-build-core in https://github.com/pytorch/pytorch/pull/180247. I think our PR adding force-include might be useful for that work. Can you give some examples of how this could be used to simplify that PR?

I've now got a clear picture. The most relevant file is `cmake/FileMirroring.cmake` — it's essentially a hand-rolled force-include implemented in CMake. Here's my analysis, grounded in the actual PR.

## The core opportunity: `FileMirroring.cmake` ≈ a manual `wheel.force-include`

That whole file exists to copy static source files (YAMLs, vendored headers, vendored `.py`) from places like `third_party/`, `aten/`, and `tools/` into wheel destinations. That is exactly what `[tool.scikit-build.wheel.force-include]` does — but in `pyproject.toml`, with no CMake plumbing, no `SKBUILD_PLATLIB_DIR` shim, and no standalone-cmake guard.

### Example 1 — single-file mirrors → `wheel.force-include`

The valgrind headers and the torchgen YAMLs are textbook cases.

**Before** (`cmake/FileMirroring.cmake`):
```cmake
install(FILES
  "${PROJECT_SOURCE_DIR}/aten/src/ATen/native/native_functions.yaml"
  "${PROJECT_SOURCE_DIR}/aten/src/ATen/native/tags.yaml"
  DESTINATION "${SKBUILD_PLATLIB_DIR}/torchgen/packaged/ATen/native")

install(FILES "${PROJECT_SOURCE_DIR}/third_party/valgrind-headers/callgrind.h"
  DESTINATION "${SKBUILD_PLATLIB_DIR}/torch/utils/benchmark/utils/valgrind_wrapper")
install(FILES "${PROJECT_SOURCE_DIR}/third_party/valgrind-headers/valgrind.h"
  DESTINATION "${SKBUILD_PLATLIB_DIR}/torch/utils/benchmark/utils/valgrind_wrapper")
```

**After** (`pyproject.toml`):
```toml
[tool.scikit-build.wheel.force-include]
"aten/src/ATen/native/native_functions.yaml" = "torchgen/packaged/ATen/native/native_functions.yaml"
"aten/src/ATen/native/tags.yaml"             = "torchgen/packaged/ATen/native/tags.yaml"
"third_party/valgrind-headers/callgrind.h"   = "torch/utils/benchmark/utils/valgrind_wrapper/callgrind.h"
"third_party/valgrind-headers/valgrind.h"    = "torch/utils/benchmark/utils/valgrind_wrapper/valgrind.h"
```

The `cutedsl` copy is the same pattern, and force-include's destination is a full path so the CMake `RENAME` is free:
```toml
"third_party/cutlass/examples/python/CuTeDSL/blackwell/grouped_gemm.py" = "torch/_inductor/kernel/vendored_templates/cutedsl/kernels/cutedsl_grouped_gemm.py"
```
Force-include is always-strict (missing source → `FileNotFoundError`), which matches their existing `if(EXISTS ...) … else FATAL_ERROR` guard — though they'd lose the bespoke "run git submodule update" hint, so that one's a judgment call.

### Example 2 — whole-directory mirror → one `wheel.force-include` entry

`wheel.force-include` accepts a directory and recurses (skipping VCS/`__pycache__`):

**Before:**
```cmake
install(DIRECTORY "${PROJECT_SOURCE_DIR}/aten/src/ATen/templates/"
  DESTINATION "${SKBUILD_PLATLIB_DIR}/torchgen/packaged/ATen/templates")
```
**After:**
```toml
"aten/src/ATen/templates" = "torchgen/packaged/ATen/templates"
```

### Example 3 — "keep one file from an otherwise-excluded tree" → exclude broadly + force-include

This is where the **"force-includes survive `wheel.exclude`"** fix I just landed pays off directly. Today the PR does:
```toml
exclude = [
    "torch/csrc/**",
    "!torch/csrc/inductor/aoti_runtime/model.h",   # gitignore-negation re-include
    ...
]
```
and the comment notes `model.h` is *also* "staged explicitly via cmake." With force-include you express the intent once and it's guaranteed to win over the broad exclude:
```toml
exclude = ["torch/csrc/**", ...]   # no negation needed

[tool.scikit-build.wheel.force-include]
"torch/csrc/inductor/aoti_runtime/model.h" = "torch/csrc/inductor/aoti_runtime/model.h"
```
That drops both the `!` re-include and a CMake staging rule for the same file.

## A bonus correctness angle: editable installs

PyTorch uses `editable.mode = "redirect"`. Force-included files are baked into the editable wheel too (that was part of this feature's design), so these mirrored files (e.g. the valgrind headers) become present and consistent in editable installs — whereas `install(FILES … DESTINATION SKBUILD_PLATLIB_DIR)` rules behave differently under redirect editable layout. So this isn't only fewer lines; it removes an editable/non-editable discrepancy.

## Where force-include is *not* the right tool (so the PR keeps these as-is)

I want to be honest about the limits, because two patterns in the PR look like candidates but aren't:

- **Extension-filtered "keep" negations** — `!torch/nativert/**/*.py`, `!torch/lib/libshm/**/*.h`. Force-include of a directory copies *every* file (it can't filter to just `.py`/`.h`), and because forced files now intentionally override `wheel.exclude`, you can't trim them afterward either. Leave these as gitignore negations.
- **Directory copy with `PATTERN … EXCLUDE`** — the `tools/autograd/` mirror excludes `BUILD.bazel`/`*.bzl`. Same reason: force-include can't do per-pattern exclusion within the copied tree. Keep it in CMake (or pre-filter), don't force-include it.

Also note: all these destinations are platlib-relative (`torch/…`, `torchgen/…`), so they need **no** leading-slash wheel-tree syntax and thus don't depend on `experimental` for force-include itself — though PyTorch already sets `experimental = true` anyway.

---

Net: the strongest, safest simplification is **migrating the static `install(FILES/DIRECTORY …)` rules in `FileMirroring.cmake` into `[tool.scikit-build.wheel.force-include]`** (Examples 1–2), which could let most of that file disappear, plus **the single-file `model.h` keep** (Example 3). The big `exclude` allowlist and the pattern-filtered copies should stay as they are.

Want me to draft the exact `pyproject.toml` block and the corresponding `FileMirroring.cmake` trim as a concrete patch against their PR branch?

henryiii:
Based on that analysis, I'm changing directories to support exclusion, with only exact files overriding exclusion. That frees up several more to also work correctly here.

henryiii:
@LecrisUT I iterated quite a bit on this, but I eventually went with your design; there's an `sdist.force-include` and `wheel.force-include`. This is also quite a bit like hatchling (though hatchling also has a build.force-include as a shortcut for both). Usually `sdist.force-include` should be enough for pulling in other files, and `wheel.force-include` can put a file from the sdist (including one added) into the wheel anywhere (I added explicit chaining support).

I don't have a way to enable/disable the error if a file is missing. Unlike hatchling, we do know if we are an sdist or a source build, so sdist builds don't try to re-copy the missing file, so I think strictness was a better default. I think people could probably use overrides if they had something actually optional. Which I assume is pretty rare.

There's no way to customize the error message, so PyTorch would lose a custom error message going this way.

Currently, this is just a copy for editable installs, though I should be able to add a mapping and use that as a followup.

docs/configuration/index.md
```diff
@@ -359,6 +359,85 @@ assume `wheel.platlib = false` (purelib targeted instead).
...
+(`wheel.exclude` for wheels, `sdist.exclude` for SDists): naming an exact source
+is an explicit request, so it wins even if an exclude pattern matches its
+destination. A force-included _directory_ stays subject to that exclude, so a
+bulk tree copy can still be trimmed by an exclude pattern (e.g. force-include a
+directory and exclude `**/*.bzl` to drop the Bazel files from it).
+
+Both tables can be set through any config source, including config-settings and
+the `SKBUILD_SDIST_FORCE_INCLUDE` / `SKBUILD_WHEEL_FORCE_INCLUDE` environment
+variables.
+
```
henryiii:
```suggestion
```

Technically true, but not sure we should be advertising this as it's a pretty niche feature.

docs/configuration/index.md
```diff
@@ -359,6 +359,85 @@ assume `wheel.platlib = false` (purelib targeted instead).
...
+and is used directly. When it is built from the source tree (or an editable
+install) the file was never materialized; a `wheel.force-include` source missing
+on disk is then resolved through `sdist.force-include` (by exact destination, or
+under a force-included directory) and read from that original source instead. An
+on-disk file always wins, so the vendored copy is preferred when present.
+
+For cases the automatic resolution cannot express — e.g. the wheel source is the
+_original_ external path rather than the SDist output — use
+[overrides](#overrides) keyed on `from-sdist`. Because override dicts merge
+rather than replace, keep the external entry in the `from-sdist = false`
```
henryiii: I think this is only if inherit is set on this table, will check.

LecrisUT:
> @LecrisUT I iterated quite a bit on this, but I eventually went with your design; there's an `sdist.force-include` and `wheel.force-include`.

Did I have a design on this? :eyes: :grin: I probably commented about using the overloads to avoid an extra layer of namespacing, but I guess that did not work? I think it's fine with either the hatchling-like approach or our overrides overall, although we already have the `sdist` and `wheel` namespaces, so should probably be more intuitive with that.

henryiii:
You suggested using hatchling style `wheel.force-include` and `sdist.force-include` instead of the inline tables. Overrides can be used, but sdist and wheel are quite different (sdist is mapping source to an SDist location, wheel is mapping an sdist to a wheel location), they interact with different exclusions (sdist and wheel), etc. So overrides are useful (there's one example in the docs), but having two is still best I think.

Issue #1367, opened 2026-06-25
PyTorch pain points (automated)


henryiii:
I pointed Claude at https://github.com/pytorch/pytorch/pull/180247 and asked for a report of pain points that we could possibly improve. Here's the report:

:robot: _AI text below_ :robot:

I've now read the full migration surface. The pyproject.toml comments are essentially a field log of every place scikit-build-core forced a workaround, which makes this easy to mine. Here's what I found, framed as scikit-build-core feature opportunities and ordered roughly by impact-vs-effort.

## The big one: wheel file selection (`tool.scikit-build.wheel.exclude`)

`pyproject.toml:145-271` is **~130 lines of gitignore-style exclude globs** whose only job is to claw the wheel back down to the legacy `setup.py` surface. The comment admits it's a stopgap: *"This block is intended as a temporary filter; a follow-up will replace it with an install-component-based scheme that suppresses third-party install rules at source."*

The root cause: scikit-build-core ships **everything** the CMake `install()` rules emit plus everything the package-walk finds (~2300 extra files: `.a` archives — `libtorch_xpu_ops.a` alone is ~1.1 GB — `lib64/` cmake configs, third-party `share/cmake/*`, `*.lib` import libs up to ~1.3 GB for `dnnl.lib`, source trees, etc.).

Feature opportunities, easiest first:
- **Sensible default excludes for wheels**: static archives (`*.a`), `lib64/` cmake-config dirs, `share/cmake/<thirdparty>` almost never belong in a Python wheel. A default-on `wheel.exclude-build-artifacts` would erase a big chunk of this block.
- **CMake install-component selection** (`install.components = [...]`): the *real* fix they want — only install the components torch declares, so third-party subproject `install()` rules never fire. This is the structural answer and is already a known scikit-build-core gap.
- **Allowlist / include mode** as an alternative to exclude — the legacy `package_data` was an allowlist (`lib/*.so*`), which is far shorter to express than its negation.

## Compiler-detection env leak (4 override blocks)

`pyproject.toml:62-103` — four separate overrides exist solely because scikit-build-core's `set_environment_for_gen()` picks up Python's `sysconfig` `CC`/`CXX` and injects it into the CMake subprocess env, which then:
- leaks conda's `-B .../compiler_compat` narrow sysroot → breaks `try_compile` probes (FindBLAS `sgemm_`, FindOpenMP),
[...]

PR #1369, opened 2026-06-25
fix: cap SKBUILD_PROJECT_VERSION to four components

henryiii:

Caps `SKBUILD_PROJECT_VERSION` to at most four dot-separated release components (`major.minor.patch.tweak`) so it is always valid for CMake's strict `project(VERSION ...)` parser. This addresses #1367, where projects currently hand-strip PEP 440 suffixes.

Previously `SKBUILD_PROJECT_VERSION` was release-only but not guaranteed CMake-valid: `version.release` can have more than four components (`(1,2,3,4,5)` -> invalid). It is now truncated via `version.release[:4]`.

No zero-padding is added. CMake accepts up to four non-negative integer segments and is happy with a bare one- or two-component version, so `2` and `2.6` are left as-is. The `release` segment is already free of epoch/pre/post/dev/local. The full, unmodified version remains available as `SKBUILD_PROJECT_VERSION_FULL`, so no new variable is needed.

Examples:
- `2` -> `2`
- `2.6` -> `2.6`
- `1.2.3.4` -> `1.2.3.4`
- `1.2.3.4.5` -> `1.2.3.4`
- epoch/pre/post/dev/local are already stripped by `version.release`, e.g. `1!2.3.4` -> `2.3.4`, `1.0.0a1` -> `1.0.0`, `1.2.3.dev4+local` -> `1.2.3`

### Compatibility note

This changes the existing variable rather than adding a new one. Only versions with 5+ release components now render differently in `SKBUILD_PROJECT_VERSION`. Anyone needing the exact value should use `SKBUILD_PROJECT_VERSION_FULL`.

## Test plan
- `tests/test_builder.py::test_builder_project_version_cmake` is parametrized over the edge cases, asserting both the capped `SKBUILD_PROJECT_VERSION` and the untouched `SKBUILD_PROJECT_VERSION_FULL` in the generated init cache.
- `prek -a --quiet` clean (ruff + mypy + typos).
- Docs: updated the `SKBUILD_PROJECT_VERSION` entry in `docs/guide/cmakelists.md`.

### DIFF
--- a/src/scikit_build_core/builder/builder.py
+++ b/src/scikit_build_core/builder/builder.py
@@ -226,9 +226,7 @@ class Builder:
             canonical_name = name.replace("-", "_").replace(".", "_")
             cache_config["SKBUILD_PROJECT_NAME"] = canonical_name
         if version is not None:
-            # Cap to four components so it is valid for project(VERSION ...),
-            # which accepts up to major.minor.patch.tweak. No padding is needed;
-            # SKBUILD_PROJECT_VERSION_FULL retains the full version.
+            # Cap to four components so it is valid for project(VERSION ...)
             cache_config["SKBUILD_PROJECT_VERSION"] = ".".join(
                 str(v) for v in version.release[:4]
             )
--- a/src/scikit_build_core/builder/builder.py
+++ b/src/scikit_build_core/builder/builder.py
@@ -226,8 +226,11 @@ class Builder:
             canonical_name = name.replace("-", "_").replace(".", "_")
             cache_config["SKBUILD_PROJECT_NAME"] = canonical_name
         if version is not None:
+            # Cap to four components so it is valid for project(VERSION ...),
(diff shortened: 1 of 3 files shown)


henryiii:
I think we should probably just cap the regular version one to three digits instead. We have a full version for more.

src/scikit_build_core/builder/builder.py
```diff
@@ -226,8 +226,11 @@ def configure(
             canonical_name = name.replace("-", "_").replace(".", "_")
             cache_config["SKBUILD_PROJECT_NAME"] = canonical_name
         if version is not None:
+            # Cap to four components so it is valid for project(VERSION ...),
+            # which accepts up to major.minor.patch.tweak. No padding is needed;
+            # SKBUILD_PROJECT_VERSION_FULL retains the full version.
```
henryiii:
```suggestion
            # Cap to four components so it is valid for project(VERSION ...)
```

PR #1372, opened 2026-06-25
feat: honor CMAKE_BUILD_TYPE from the environment

henryiii:

Part of #1367.

## Problem

CMake 3.22+ reads `CMAKE_BUILD_TYPE` from the environment, but only as a *default* for the first configure — an explicit `-DCMAKE_BUILD_TYPE` on the command line overrides it. Since scikit-build-core's `cmake.build-type` defaults to `"Release"`, it **always** passes `-DCMAKE_BUILD_TYPE=Release`, so an env-set `CMAKE_BUILD_TYPE=RelWithAssert` is silently ignored. (This is one of the pain points in #1367 — pytorch needed extra override blocks to work around it.)

## Fix

After settings are resolved, if `CMAKE_BUILD_TYPE` is set in the environment **and** `cmake.build-type` was not explicitly configured (via TOML, `SKBUILD_CMAKE_BUILD_TYPE`, config-settings, or an override), use the environment value.

Resulting precedence:
1. Explicit `cmake.build-type` — wins
2. `CMAKE_BUILD_TYPE` environment variable
3. `"Release"` default

This is done in Python rather than relying on CMake's native env support, so it works on CMake versions older than 3.22 (scikit-build-core supports down to 3.15) and uses the same `env` mapping the rest of the settings reader uses.

## Tests

- Added a regression test confirming `CMAKE_BUILD_TYPE` is honored when `build-type` is unset (verified it fails without the fix).
- Added a test confirming an explicit `cmake.build-type` still beats `CMAKE_BUILD_TYPE`.
- Docs note added under "Configuring CMake arguments and defines".

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -428,6 +428,12 @@ You can select a different build type, such as `Debug`:
 
 ```
 
+If `cmake.build-type` is left at its default and `CMAKE_BUILD_TYPE` is set in
+the environment, that value is used instead. This lets you override the build
+type without editing `pyproject.toml` (for example
+`CMAKE_BUILD_TYPE=RelWithDebInfo`), mirroring CMake's own handling of the
+variable.
+
 You can specify CMake defines as strings or bools:
 
 ````{tab} pyproject.toml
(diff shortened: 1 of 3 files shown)


henryiii:
@claude rebase

Issue #822, opened 2024-07-23
How to use gcc compiler on Github windows runners with microsoft visual studio installed


jiawlu:
I'm using scikit-build-core to build my c++ program on github windows runners. I would like to use gcc instead of microsoft visual studio as the compiler. Below is how I configure compiler settings in my program, but it does not work (see the [log](https://github.com/jiawlu/OSM2GMNS/actions/runs/10062279697/job/27814408535), MSVS was used)

Github workflow file
```
name: build-wheels

on:
  push:
    branches: [main]
  pull_request:
    branches: [main, release]
  workflow_dispatch:

jobs:
  build_wheels:
    name: Build wheel for ${{ matrix.buildplat[1] }}
    runs-on: ${{ matrix.buildplat[0] }}
    strategy:
      fail-fast: false
      matrix:
        buildplat:
          - [windows-2022, win_amd64]
        python: ["cp38"]
    steps:
      - uses: actions/checkout@v3

      - name: Build wheels for win
        if: ${{ contains(matrix.buildplat[1], 'win') }}
        uses: pypa/cibuildwheel@v2.19
        env:
          CIBW_BUILD: ${{ matrix.python }}-${{ matrix.buildplat[1] }}
          CIBW_ENVIRONMENT_WINDOWS: >
            CC=gcc
            CXX=g++
            SKBUILD_CMAKE_ARGS="-DCMAKE_C_COMPILER=gcc;-DCMAKE_CXX_COMPILER=g++"

      - uses: actions/upload-artifact@v3
        with:
          path: ./wheelhouse/*.whl
```

pyproject.toml file

```
[project.urls]
"Source Code" = "https://github.com/jiawlu/OSM2GMNS"
"Bug Tracker" = "https://github.com/jiawlu/OSM2GMNS/issues"

[build-system]
requires = ["scikit-build-core>=0.9.0", "ninja>=1.0"]
build-backend = "scikit_build_core.build"

[project]
name = "osm2gmns"
version = "1.0.0dev1"
description = "convert map data from OpenStreetMap to network files in GMNS format"
authors = [{ name = "Jiawei Lu", email = "lujiaweiwk@gmail.com" }, { name = "Xuesong (Simon) Zhou", email = "xzhou74@asu.edu" }]
readme = "README.rst"

requires-python = ">=3.8"
dependencies = []

classifiers = [
  "License :: OSI Approved :: GNU General Public License v3 or later (GPLv3+)",
[...]

PR #1377, opened 2026-06-27
feat: add env table for build environment variables

henryiii:
This is very useful for PyTorch (avoiding a wrapper around us), but also solves several other issues. Also folded in #1370 using this mechanism.

## Summary

Adds a top-level `[tool.scikit-build.env]` table that sets environment variables for the CMake configure, build, and install subprocesses. This gives a config-native way to control things CMake/the generator read *from the environment* (`CMAKE_BUILD_PARALLEL_LEVEL`, `CC`/`CXX`, `CFLAGS`, `CMAKE_PREFIX_PATH`, compiler launchers, …) — for `-D` cache entries, `cmake.define` is still the tool.

Motivated by pytorch/pytorch#180247, which currently wraps scikit-build-core with a custom PEP 517 backend solely to map `MAX_JOBS` → `CMAKE_BUILD_PARALLEL_LEVEL`. With this table that wrapper can be deleted:

```toml
[tool.scikit-build.env]
CMAKE_BUILD_PARALLEL_LEVEL = { env = "MAX_JOBS" }
```

Also addresses the `CMAKE_PREFIX_PATH`/compiler pieces of #1367, and gives a config-native path for #822 (selecting a compiler).

## Semantics

Each value is a literal string, or a table with `env` (read another variable), `default` (fallback) / explicit literal), and `force`.

- **setdefault by default** — a variable already set in the environment is preserved.
- **`force = true`** — overwrite (CMake `set(... FORCE)` analogue).
- **skip when unresolved** — an unset source `env` with no `default` leaves the variable unset (so no `MAX_JOBS` → no `CMAKE_BUILD_PARALLEL_LEVEL`).

Works in `pyproject.toml`, config-settings (`-Cenv.FOO=bar`), `SKBUILD_ENV`, and `[[tool.scikit-build.overrides]]` bodies (merges per-key). The table form (`{ env = … }`, `force`) is TOML-only, matching `cmake.define`'s `EnvVar` form. It is independent of the `if.env` override *condition*, which only matches the ambient process environment.

## Folds in `cmake.use-sysconfig-compiler`

That setting is unreleased, so rather than ship it, its behavior moves into this table: listing `CC`/`CXX` in `env` suppresses the default sysconfig compiler injection (even when the entry resolves to nothing), so `CC = { env = "CC" }` defers to CMake's own detection. This is strictly more flexible — compiler choice can now be scoped per-platform/state via overrides, and `CC`/`CXX` controlled independently.

## Tests

- Settings parsing across TOML / config-settings / `SKBUILD_ENV`, including literal / `{ env, default }` / `force` forms.
- Builder application: setdefault, force, skip-when-unresolved, and the `MAX_JOBS` → `CMAKE_BUILD_PARALLEL_LEVEL` regression.
- `env_managed_keys` suppression of the sysconfig `CC`/`CXX` injection.
- Schema valid/invalid cases; generated schema/README/config-reference regenerated; docs build is warning-free.

https://claude.ai/code/session_01DRpvaLqeQuYgqoETCKEKiQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ search.site-packages = true
 # List dynamic metadata fields and hook locations in this table.
 metadata = {}
 
-# A table of environment variables to set for the CMake subprocesses. Additive.
+# A table of environment variables to set for the CMake subprocesses.
 env = {}
 
 # Strictly check all config options.
--- a/README.md
+++ b/README.md
@@ -177,9 +177,6 @@ cmake.source-dir = "."
 # Do not pass the current environment's python hints such as ``Python_EXECUTABLE``.
 cmake.python-hints = true
 
-# Set ``CC``/``CXX`` from Python's sysconfig compiler when not already set in the
-cmake.use-sysconfig-compiler = true
-
 # The versions of Ninja to allow.
 ninja.version = ">=1.5"
 
@@ -293,6 +290,9 @@ search.site-packages = true
 
 # List dynamic metadata fields and hook locations in this table.
 metadata = {}
+
+# A table of environment variables to set for the CMake subprocesses. Additive.
+env = {}
 
 # Strictly check all config options.
 strict-config = true
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -638,6 +638,86 @@ You can pass raw arguments directly to the build tool, as well:
 
 ```
 
+## Environment variables for the build
+
+The `[tool.scikit-build.env]` table sets environment variables for the CMake
+configure, build, and install subprocesses. Use it for things CMake or the
+generator read _from the environment_ — `CC`/`CXX`, `CFLAGS`,
+`CMAKE_PREFIX_PATH`, compiler launchers, parallel-build level, and so on. For
+CMake `-D` cache entries, use `cmake.define` instead.
+
+Each value is a literal string, or a table that reads from another environment
+variable (`{ env = "OTHER", default = "..." }`). By default a variable is only
+set if it is not already present (`setdefault`); add `force = true` to overwrite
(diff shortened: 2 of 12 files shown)


henryiii:
GPT 5.5 in Codex:

The new env table accepts malformed runtime settings and can change overwrite behavior incorrectly. This should be fixed before considering the patch correct.

Review comment:

- [P2] Validate env table value types before coercing — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/settings/skbuild_model.py:94-96
  When a project sets an invalid env table like `FOO = { default = "x", force = "false" }`, the build backend does not run the JSON schema and this constructor silently coerces `force` with `bool(...)`, so the string `"false"` becomes `True` and can unexpectedly overwrite an existing environment variable. Other settings reject wrong TOML types during conversion, and these `env`/`default`/`force` fields should do the same instead of relying only on schema validation.

henryiii:

Fixed in 1e9ab03. `EnvValue.__init__` now validates types at conversion time instead of coercing:

- `force` must be a real TOML boolean — `force = "false"` (or `0`/`1`) now raises a `TypeError` rather than `bool("false")` → `True` silently enabling overwrite. (`bool` is an `int` subclass, so the `isinstance(force, bool)` check rejects ints too.)
- `env` and `default` must be strings.

This matches how other settings reject bad TOML types during conversion (surfaced as the usual `ExceptionGroup`), so it no longer relies on the JSON schema being run. Added a regression test (`test_env_value_rejects_bad_types`) that fails on the old `bool(...)` coercion, plus an end-to-end `pyproject.toml` case (`test_skbuild_settings_env_table_bad_force_rejected`).

PR #1379, opened 2026-06-27
feat: add ${SKBUILD_<TREE>_DIR} prefix for wheel-tree targeting

henryiii:

## Summary

Adds a non-experimental `${SKBUILD_<TREE>_DIR}/...` prefix syntax for targeting special wheel trees in `wheel.install-dir` and `wheel.force-include`, mirroring the `SKBUILD_*_DIR` CMake cache variables. This replaces the confusing leading-slash "absolute path" form (`/data`, `/scripts`, ...) — which *looks* like a filesystem absolute path but isn't — with one consistent vocabulary shared between the config side and CMakeLists.

The old `/data/`-style form keeps working unchanged and stays gated behind `experimental = true`; the new form is **not** experimental.

```toml
[tool.scikit-build]
wheel.install-dir = "${SKBUILD_DATA_DIR}/mypackage"
wheel.force-include = { "tools/run.sh" = "${SKBUILD_SCRIPTS_DIR}/run.sh" }
```

Available trees: `${SKBUILD_PLATLIB_DIR}` (default), `${SKBUILD_PURELIB_DIR}`, `${SKBUILD_DATA_DIR}`, `${SKBUILD_HEADERS_DIR}`, `${SKBUILD_SCRIPTS_DIR}`, `${SKBUILD_METADATA_DIR}`, `${SKBUILD_NULL_DIR}`.

## Changes

- **`build/_pathutil.py`** — `resolve_wheel_tree()` matches a `${SKBUILD_<TREE>_DIR}` prefix and resolves it through the same path as the leading-slash form (same platlib/purelib aliasing, same valid-tree check, same `..` traversal rejection). Not experimental-gated. Single chokepoint, so wheel build / force-include / hatch plugin are all covered.
- **`build/_editable.py`** — the rebuild-incompatibility check now also rejects a `${SKBUILD_...}` install-dir (a rebuild can't target a tree above platlib).
- **Docs + model docstrings** — present the var form as preferred, leading-slash as deprecated; regenerated `configs.md`/`README` via `nox -t gen`. Documented the CLI single-quoting caveat (the shell would otherwise expand `${SKBUILD_DATA_DIR}` when passed through `config-settings`).
- **Tests** — integration tests (incl. one asserting the var form works with `experimental = false`), direct `resolve_wheel_tree` unit tests (all trees, no-remainder, purelib alias, bad name, traversal, non-matching literal), and an editable rebuild-rejection test.

## Notes / open question

The editable rebuild rejection is intentionally **coarse**: it rejects *any* `${SKBUILD_...}` install-dir, including `${SKBUILD_PLATLIB_DIR}/...` which technically stays within platlib. This matches the pre-existing behavior for `/platlib`. Could be refined to allow rebuild when the target tree is platlib/purelib — happy to do so if preferred.

## Verification

- `prek -a --quiet` clean (includes mypy)
- `tests/test_force_include.py`, `tests/test_editable_unit.py`, `tests/test_editable.py`, `tests/test_pyproject_pep517.py` pass

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -333,10 +333,10 @@ This matches the cache variables available from within CMake.
 
 When passing this through PEP 517 `config-settings` on a command line, quote it
 so the shell does not expand `${SKBUILD_DATA_DIR}` as an environment variable
-(e.g. `--config-settings='wheel.install-dir=${SKBUILD_DATA_DIR}/mypackage'`).
+(e.g. `-C 'wheel.install-dir=${SKBUILD_DATA_DIR}/mypackage'`).
 
 The older leading-slash spelling (`/data`, `/scripts`, ...) selects the same
-trees but is still gated behind `experimental = true`.
+trees but is gated behind `experimental = true`, and deprecated.
 
 :::
 
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -316,11 +316,27 @@ For example, to mimic scikit-build classic:
 wheel.install-dir = "mypackage"
 ```
 
+You can target a different wheel tree by prefixing the install dir with the
+matching `SKBUILD_*_DIR` CMake variable name:
+
+```toml
+[tool.scikit-build]
+wheel.install-dir = "${SKBUILD_DATA_DIR}/mypackage"
+```
+
+The available trees are `${SKBUILD_PLATLIB_DIR}` (the default),
+`${SKBUILD_PURELIB_DIR}`, `${SKBUILD_DATA_DIR}`, `${SKBUILD_HEADERS_DIR}`,
+`${SKBUILD_SCRIPTS_DIR}`, `${SKBUILD_METADATA_DIR}`, and `${SKBUILD_NULL_DIR}`.
+This matches the cache variables available from within CMake.
+
 :::{warning}
 
-You can select a different wheel target directory, as well, but that syntax is
-experimental; install to `${SKBUILD_DATA_DIR}`, etc. from within CMake instead
-for now.
+When passing this through PEP 517 `config-settings` on a command line, quote it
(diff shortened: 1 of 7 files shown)


henryiii:

Pushed `8828a92` addressing both the typo-guard request and Codex's two P2s:

- **Typo guard** — a leading `${...}` that isn't a recognized `${SKBUILD_<TREE>_DIR}` variable is now a hard error (e.g. `${SKBUILD_DTA_DIR}`, `${HOME}`, unterminated `${...`), instead of silently becoming a literal filename. A `${...}` *later* in the path is still treated as an ordinary path component, as discussed.
- **P2 – escaping remainders** (`_pathutil.py`) — the remainder after a tree selector is rejected if it is absolute, drive-qualified, or contains a backslash (`${SKBUILD_DATA_DIR}//pkg` → `/pkg`, `${SKBUILD_DATA_DIR}/C:/x`, ...). This is enforced in `resolve_wheel_tree` so both `wheel.install-dir` and `wheel.force-include` are covered, and it also closes the same pre-existing hole in the leading-slash form (`/data//pkg`).
- **P2 – editable rebuild** (`_editable.py`) — instead of rejecting any `${...}`/`/` install-dir, it now only rejects ones that resolve to a *non-platlib* tree. A `${SKBUILD_PLATLIB_DIR}/pkg` (or `${SKBUILD_PURELIB_DIR}` / `/platlib`) selector is reduced to its platlib-relative remainder (`pkg`) and that normalized value is passed to the redirect shim, which joins it onto the platlib root and only understands relative paths.

New tests cover the typo rejections, the not-leading literal case, escaping-remainder rejection (both forms), and that a platlib/purelib selector is allowed with rebuild and reaches the shim as the reduced remainder.

`prek -a` clean; `test_force_include.py`, `test_editable_unit.py`, `test_editable.py` pass.

PR #1381, opened 2026-06-27
fix: use `Path.as_posix()` for including generated files in sdist/wheel

shiftinv:
Heya! I noticed that our CI seemed to be failing after adding an auto-generated `_version.py` file ([config](https://github.com/DisnakeDev/dave.py/blob/409901d0337ea37b35bf0d3d6c54574c0bb24abe/pyproject.toml#L153-L159), as outlined in the [docs](https://scikit-build-core.readthedocs.io/en/stable/configuration/dynamic.html#generate-files-with-dynamic-metadata)), but weirdly only on Windows, which didn't seem to find the module while running tests from the built wheel.
As it turns out, gitignore files (and, by extension, `pathspec`) exclusively support forward slashes, whereas the `str(gen.path)` call would turn the forward slashes from the config into the windows backslash path separator, which meant the generated file wasn't being added to the wheel in the first place. Using `.as_posix()` instead avoids that.

I hope the test setup is okay-ish and in line with the rest of the project, I mostly just took inspiration from `test_pyproject_purelib.py` and `test_force_include.py` c:

### DIFF
--- a/src/scikit_build_core/build/sdist.py
+++ b/src/scikit_build_core/build/sdist.py
@@ -128,7 +128,7 @@ def build_sdist(
         if gen.location == "source":
             contents = generate_file_contents(gen, metadata)
             gen.path.write_text(contents, encoding="utf-8")
-            settings.sdist.include.append(str(gen.path))
+            settings.sdist.include.append(gen.path.as_posix())
 
     sdist_dir.mkdir(parents=True, exist_ok=True)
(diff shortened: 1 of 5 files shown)


henryiii:
Thanks! I'm dropping the compiled marker since this doesn't compile anything. I verified it's quite fast (0.11 s for me), so fine to keep.

Issue #1378, opened 2026-06-27
abi3t wheel tag from main is not accepted by current pip/packaging


honglei:
Hi, thanks for adding abi3t support in #1269.

I searched existing issues before opening this. I found:

- #1252, closed by #1269
- pypa/pip#13879, closed after pip vendored packaging with PEP 803 support
- several project-side tracker issues for publishing abi3t wheels, such as neutrinoceros/rlic#436 and pola-rs/polars#27955

I did not find an existing scikit-build-core issue for the specific behavior below.

## Reproducer

I tested scikit-build-core `main` at commit `e4fe16c5a1206bd025424c195f7c383da6264080` on CPython 3.15.0b3 free-threaded:

```text
scikit-build-core 0.12.3.dev89+ge4fe16c5a
cmake 4.3.4
pip 26.1.2
packaging 26.2
Python 3.15.0b3 free-threading build
```

Build command:

```bash
python -m pip wheel . -w /tmp/wheelhouse --no-deps --no-build-isolation \
  --config-settings=wheel.py-api=cp315t \
  --config-settings=cmake.args="-DAISCAT_ABI3T=ON;-DCMAKE_BUILD_TYPE=Release"
```

Generated wheel:

```text
aiscat-0.68.14-cp315t-abi3t-linux_x86_64.whl
Tag: cp315t-abi3t-linux_x86_64
contains: aiscat/_core.abi3t.so
```

Install result on the same interpreter:

```text
ERROR: aiscat-0.68.14-cp315t-abi3t-linux_x86_64.whl is not a supported wheel on this platform.
```

On that interpreter, `packaging.tags.sys_tags()` includes `cp315-abi3t-*` tags, but does not include `cp315t-abi3t-*`.

## Tag compatibility question

As I understand PEP 803 and the current pip/packaging behavior, installer-compatible abi3t wheel tags use a GIL-independent Python tag such as `cp315`, with the free-threaded stable ABI represented in the ABI tag:

```text
cp315-abi3t-*
cp315-abi3.abi3t-*
```

Current pip/packaging does not accept:

```text
cp315t-abi3t-*
```

So this may be a mismatch between scikit-build-core's current interpretation of `wheel.py-api=cp315t` and the PEP 803 / installer-compatible wheel tag set.

## Why this looks like a release blocker

scikit-build-core main currently documents `wheel.py-api=cp315t` and its tests appear to expect `cp315t-abi3t-*`:

```python
[...]

PR #1382, opened 2026-06-28
feat: support combined abi3.abi3t wheel tag

henryiii:
The new support in CMake 4.4rc2 seems to incorrectly assume you need free-threaded Python to build the new Stable ABI (probably due to the `t` in it). See [gitlab.kitware.com/cmake#27907](https://gitlab.kitware.com/cmake/cmake/-/work_items/27907) for info.

I want to see if I can do this a little more simply.

## Summary

Follow-up to #1378 / #1380. Lets a project broaden a free-threaded stable-ABI wheel to also cover GIL-enabled CPython, with a single build, and adds a worked example.

Setting `wheel.py-api = "cp315.cp315t"` on a **free-threaded** build emits:

```
cp315-abi3.abi3t-<platform>
```

which expands to the two installer tags `cp315-abi3` and `cp315-abi3t`.

## How this works (single build, single binary)

Per PEP 803, `abi3t` is a strict **subset** of `abi3`. A single extension compiled for `abi3t` (with `Py_TARGET_ABI3T`) is loadable on **both** free-threaded and GIL-enabled CPython 3.15+. So the combined `cp315-abi3.abi3t` wheel is produced by **one** build with **one** `.abi3t.so` — not two builds merged. `cp315t` remains the deliberately-narrow label (`cp315-abi3t`) for the identical binary.

This was verified empirically end-to-end: one `cp315-abi3.abi3t` wheel built on free-threaded 3.15t installs and runs on both GIL 3.15 and free-threaded 3.15t. (abi3t compiled against GIL *headers* also loads on both — abi3t-from-GIL is real at the C level.)

## Behavior

| `py-api` | free-threaded build | GIL build |
| --- | --- | --- |
| `cp315` | (ignored) | `cp315-abi3` |
| `cp315t` | `cp315-abi3t` | (ignored) |
| `cp315.cp315t` | `cp315-abi3.abi3t` | `cp315-abi3` |

On a GIL interpreter only `abi3` is built reliably today (CMake FindPython abi3t support needs 4.4+; Windows lacks `python3XXt.lib` on GIL installs), so `cp315.cp315t` falls back to `cp315-abi3`. Single-tag behavior is unchanged.

## Changes

- `wheel_tag.py`: on a free-threaded build, when the classic ABI is also requested, broaden the `abi3t` tag to `abi3.abi3t`. Stable-ABI resolution extracted to `_stable_abi_tag` helper.
- `builder.py`: parse dotted `py-api`, build the ABI matching the current interpreter, and derive `SKBUILD_SABI_VERSION` from the selected component (previously crashed on a dotted value).
- **New example** `docs/examples/getting_started/abi3t/`: a PEP 793 `PyModExport` module + CMakeLists, added as an ABI3t docs tab. Doc-only for now (no `main.fmf`) since the testing-farm matrix runs distro Python < 3.15.
- Docs + config reference regenerated (`nox -t gen`).

## Tests

- `test_wheel_tag_with_combined_abi3_abi3t_free_threaded` / `_gil` — combined tag on FT, `abi3` fallback on GIL.
- `test_builder_combined_abi3_abi3t` — single binary matches the interpreter (`abi3t`/`Py_TARGET_ABI3T` on FT, `abi3` on GIL), `SKBUILD_SABI_VERSION = 3.15`.

## Noted for before release (not in this PR)

Building `abi3t` currently needs CMakeLists workarounds (pass `Py_TARGET_ABI3T` as a compile define, force the SOABI suffix on CMake < 4.4) — candidates for scikit-build-core to handle so user CMakeLists stay clean.


henryiii:


The combined ABI tag logic can emit an abi3 tag for a lower CPython minor than was requested when the classic and free-threaded tags differ. That can make wheels appear compatible with interpreters the configuration did not request.

Review comment:

- [P2] Validate mixed minor versions before adding abi3 — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/builder/wheel_tag.py:184-185
  When `wheel.py-api` combines ABI requests with different minor versions, e.g. `cp316.cp315t` on CPython 3.15t, `classic_tags` is only used as a boolean here, so the result becomes `cp315-abi3.abi3t-*`. That advertises `cp315-abi3` support even though the classic ABI request was for 3.16+, making the wheel installable on unsupported GIL interpreters. Please validate matching minors or honor the classic tag's version before adding `abi3`.

henryiii:

Addressed in 7282be1. A combined `cp3XX-abi3.abi3t` wheel shares a single minor version (the free-threaded one), so `_stable_abi_tag` now requires the classic abi3 minor to be no newer than the free-threaded minor. `cp316.cp315t` on 3.15t raises `AssertionError` rather than silently emitting `cp315-abi3` (which would advertise GIL abi3 support below the requested 3.16); `cp314.cp315t` and `cp315.cp315t` still produce `cp315-abi3.abi3t`. Added a regression test (`test_wheel_tag_combined_rejects_higher_classic_minor`) and documented the constraint in the `py-api` reference.

henryiii:
Only thing left is what to do on 315t for `cp315.cp315t` (or even `cp315t` by itself). Currently, I think it will follow CMake 4.4's pattern of requiring a free-threaded interpreter to make the abi3t wheel, but that's not actually correct, you can build it from the GIL interpreter too (it's really more of an abi4, but Python programmers are allergic to making 3s into 4s). We could leave it like this for now, then introduce a new value or extra setting or something to support making the abi3t from GIL Python, probably following CMake.

PR #1389, opened 2026-06-29
feat(wheel): reproducible wheels (closes #62)

### LINKED ISSUE #62: Reproducible builds

henryiii:
I need to edit all the mentions of "changing some wheels".

🤖 _AI text below_ :robot:

Closes #62.

SDists are already reproducible; this brings wheels in line and helps the build step produce deterministic binaries. Reproducible wheels are **opt-in** (`wheel.reproducible = true`), since normalizing permissions can change the result for some existing projects.

## What changed

- **Shared helper** — new `src/scikit_build_core/_reproducible.py` (`get_reproducible_epoch`, `normalize_file_permissions`, `MIN_TIMESTAMP`), at the package root so both `build/` and `builder/` can import it. `build/sdist.py` now reuses it instead of its own copies (behavior unchanged).
- **Wheel archive** (`build/_wheelfile.py`) — `WheelWriter` gains a `reproducible` flag (default `false`):
  - `timestamp()` falls back to the fixed epoch (`1667997441`, honoring `SOURCE_DATE_EPOCH` if set) instead of each file's `st_mtime`/`time.time()`.
  - `write()` normalizes permissions to 644/755 instead of copying raw on-disk bits.
- **Setting** — new `wheel.reproducible` (default `false`, opt-in) mirroring `sdist.reproducible`; schema and config reference regenerated.
- **Build env** (`builder/builder.py`) — when enabled, `setdefault`s `SOURCE_DATE_EPOCH` into the CMake environment, so compilers that honor it (recent GCC/Clang) can emit deterministic `__DATE__`/`__TIME__`. An explicit value (shell or `tool.scikit-build.env`) always wins. Full binary reproducibility still needs compiler flags like `-ffile-prefix-map` from the user — that part is out of our control.

## Tests

- Unit: fixed-epoch timestamp and permission normalization, with a POSIX-only split for the executable/raw-mode cases (`test_wheelfile_utils.py`).
- Build env: four `SOURCE_DATE_EPOCH` export cases incl. user-env-table-wins (`test_builder.py`).
- End-to-end: build a compiled package twice with `wheel.reproducible = true` and assert identical archive metadata (`test_simplest_c.py`).

## Notes

- `wheel.reproducible` defaults to `false` (opt-in); `sdist.reproducible` remains `true`.
- The local `check-sdist` hook flags only pre-existing untracked files (`uv.lock`, `.claude/settings.local.json`), unrelated to this change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -242,6 +242,9 @@ wheel.build-tag = ""
 # Force-include files into the wheel.
 wheel.force-include = {}
 
+# Try to build a reproducible wheel.
+wheel.reproducible = true
+
 # If CMake is less than this value, backport a copy of FindPython.
 backport.find-python = "3.26.1"
 
--- a/README.md
+++ b/README.md
@@ -242,8 +242,8 @@ wheel.build-tag = ""
 # Force-include files into the wheel.
 wheel.force-include = {}
 
-# Try to build a reproducible wheel.
-wheel.reproducible = true
+# Try to build a reproducible wheel. Opt-in, as normalizing permissions can
+wheel.reproducible = false
 
 # If CMake is less than this value, backport a copy of FindPython.
 backport.find-python = "3.26.1"
--- a/README.md
+++ b/README.md
@@ -242,7 +242,7 @@ wheel.build-tag = ""
 # Force-include files into the wheel.
 wheel.force-include = {}
 
-# Try to build a reproducible wheel. Opt-in, as normalizing permissions can
+# Try to build a reproducible wheel.
 wheel.reproducible = false
 
 # If CMake is less than this value, backport a copy of FindPython.
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -514,14 +514,15 @@ And you can turn off binary stripping:
 
 ```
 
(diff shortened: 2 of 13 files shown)


henryiii:
GPT 5.5:

The new reproducible wheel option does not normalize all archive member permissions, leaving generated dist-info entries at 0664. This makes the feature's documented behavior incomplete.

Review comment:

- [P2] Normalize permissions for generated wheel entries too — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/_wheelfile.py:232-235
  When `wheel.reproducible=true`, this normalization only applies to files added through `write()`. The generated `.dist-info` files (`METADATA`, `WHEEL`, `RECORD`, entry points, and extra metadata) are added through `writestr(str, ...)`, which still hard-codes `0o664`, so reproducible wheels end up with mixed/non-normalized permissions despite the setting documenting normalized archive permissions.

PR #1390, opened 2026-06-29
feat: support dynamic-metadata 0.3 spec

henryiii:

Adds support for the standard [dynamic-metadata 0.3](https://dynamic-metadata.readthedocs.io) specification: the top-level `[[tool.dynamic-metadata]]` ordered array of tables. Reimplemented inline (no new dependency); the existing `tool.scikit-build.metadata` table keeps working but is now deprecated.

## What 0.3 changes vs. the old `tool.scikit-build.metadata`
- **Config**: top-level `[[tool.dynamic-metadata]]` *ordered array of tables* instead of a field-keyed table.
- **Hook**: `dynamic_metadata(settings, project) -> dict` (a `[project]` fragment, may set several fields) instead of `dynamic_metadata(field, settings, project) -> value`.
- **Resolution**: strictly sequential against a read-only project snapshot — no dependency graph, no cycles — plus PEP 808 add-only merge of static + dynamic values.
- **New**: optional `build_state(state)` hook; `module:Class` providers (instantiated); `dynamic_wheel(settings) -> dict[str, bool]`.

## Changes
- `metadata/__init__.py` — `_SCALAR_FIELDS` / `_EXTENDABLE_FIELDS` taxonomy for merge semantics.
- `builder/_load_provider.py` — `module:Class` loading, `BuildState`/`BUILD_STATES`, new protocols, `load_dynamic_metadata`, PEP 808 merge helpers, an adapter so the bundled legacy-signature plugins work in the array form, and `process_dynamic_metadata(project, entries, build_state)`. The old loader is preserved as `process_legacy_dynamic_metadata` (its lazy out-of-order resolution is still relied on by `test_template_needs`).
- `build/metadata.py` — `get_standard_metadata` runs the legacy table first (with a deprecation warning) then the new array, threading `build_state`.
- `build/wheel.py` / `build/sdist.py` — pass the real build state.
- `builder/get_requires.py` — also collects `get_requires_for_dynamic_metadata` from array entries.
- `build/__main__.py` — `project-table` CLI handles both forms.
- Docs + tests — new `[[tool.dynamic-metadata]]` section; 11 new unit tests + an end-to-end integration test.

## Backwards compatibility
- `tool.scikit-build.metadata` is fully supported, mapped through the preserved legacy loader, and emits a deprecation warning.
- The new array form is **not** gated behind `experimental` (the experimental gate existed because the *old* plugin interface was provisional; 0.3 is the stabilized standard).

## Open questions for review
1. `versionadded` is set to **1.0** (the spec names "scikit-build-core (1.0+)" as the supporting backend). Adjust if this lands in a different release.
2. The new array form is intentionally not `experimental`-gated — happy to add parity gating if preferred.

## Testing
- `prek -a` clean (ruff + mypy).
- Dynamic-metadata / requires test suites green (98 passed in the broad run).
- Smoke-tested the array form end-to-end via the `project-table` CLI (regex → template pipeline producing `version` and `requires-python`).

### DIFF
--- a/src/scikit_build_core/builder/_load_provider.py
+++ b/src/scikit_build_core/builder/_load_provider.py
@@ -395,9 +395,14 @@ def process_legacy_dynamic_metadata(
     """
     initial = {f: (p, c) for (f, p, c) in _load_dynamic_metadata(metadata)}
 
+    # Copy the mutable project, including the "dynamic" list that providers
+    # remove resolved fields from, so the caller's dict is left untouched.
+    project_copy = dict(project)
+    project_copy["dynamic"] = list(project_copy.get("dynamic", []))
+
     settings = DynamicPyProject(
         settings={f: c for f, (_, c) in initial.items()},
-        project=dict(project),
+        project=project_copy,
         providers={k: v for k, (v, _) in initial.items()},
     )
 
--- a/src/scikit_build_core/builder/_load_provider.py
+++ b/src/scikit_build_core/builder/_load_provider.py
@@ -8,7 +8,16 @@ import inspect
 import sys
 from collections.abc import Iterator, Mapping
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Protocol, Union, runtime_checkable
+from types import MappingProxyType
+from typing import (
+    TYPE_CHECKING,
+    Any,
+    Literal,
+    Protocol,
+    Union,
+    get_args,
+    runtime_checkable,
+)
 
 if TYPE_CHECKING:
     from collections.abc import Generator, Sequence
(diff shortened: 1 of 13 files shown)


src/scikit_build_core/builder/_load_provider.py
```diff
@@ -175,10 +383,16 @@ def __contains__(self, key: object) -> bool:
...
+def process_legacy_dynamic_metadata(
     project: Mapping[str, Any],
     metadata: Mapping[str, Mapping[str, Any]],
 ) -> dict[str, Any]:
+    """Resolve the deprecated ``tool.scikit-build.metadata`` table.
+
+    Each field names a provider; fields resolve lazily, so a provider may read
+    another field (even one defined later in the table) via ``project[...]``,
+    with circular references detected and reported.
+    """
```
LecrisUT, line 395: If deprecated, let's make it a warning?

henryiii: We generally guarantee forward compatibility. Maybe we can warn in 1.1? We do have backward compatibility; the thing I'd like to warn, and maybe remove, first would be the `experimental=True` gated stuff (custom plugins).

LecrisUT: Well it can be gated by scikit-build minimum version requested. Though that kinda defeats the purpose of a deprecation warning. Maybe we can postpone for when this becomes a PEP or equivalent standard with other backends?

henryiii: Ah, I like that idea. If we drop support for <1 someday, that would mean we could drop this too.

Issue #766, opened 2024-06-13
CMake SDist only could skip sdist step


henryiii:
Currently I don't think it's useful to use CMake only in the SDist only step, as if someone builds a wheel from source, the SDist step won't run. We should consider this use case, and see if there's a way (maybe with overrides) to add controls for Source vs. from SDist builds.

I think `if.from-sdist` would do it?

PR #1391, opened 2026-06-29
feat(sdist): add sdist.install-dir to build CMake-free wheels from an SDist (#837)

henryiii:

## Summary

Adds `sdist.install-dir`: when set (together with `sdist.cmake = true`), the SDist build runs a full configure + build + **install** instead of a configure-only run, and vendors the resulting install tree into the SDist there. A wheel built from the unpacked SDist can then repackage that tree with `wheel.cmake = false` and needs **no CMake at all**.

This closes the gap in #837 / #766: pybind11 and nanobind ship wheels that are effectively pure (headers, `*.cmake` helpers, and a few configure/install-time generated config files — no compiled extension). Today scikit-build-core runs CMake during the *wheel* build purely as a templating + install engine; this lets that generation run once in the *SDist* step instead.

```toml
[tool.scikit-build]
sdist.cmake = true
sdist.install-dir = ".cmake-install"   # upgrades the sdist CMake run to build+install, vendored here

if.from-sdist = true                    # only when building from an unpacked sdist
wheel.cmake = false                     # no CMake -> pure py3-none-any wheel
wheel.force-include.".cmake-install" = "${SKBUILD_PLATLIB_DIR}"
```

Building a wheel directly from a git checkout is unchanged (`from-sdist` is false, so the override doesn't apply and CMake runs as today). Only the SDist→wheel path becomes CMake-free. The from-sdist `force-include` targets the platlib via the non-experimental `${SKBUILD_PLATLIB_DIR}` prefix from #1379, so no `experimental = true` is needed. This branch is rebased on current `main` (which includes #1379).

## Design

- `sdist.cmake` stays a plain `bool` — on its own it is the existing configure-only run (e.g. for a FetchContent-populated source tree).
- `sdist.install-dir` (SDist-relative, default `""` = off) is the trigger that upgrades that run to configure + build + install, and names where the install tree is vendored. No magic default location.
- If `sdist.install-dir` is set without `sdist.cmake = true`, it is a no-op with a warning.

Because `sdist.cmake` remains the single "does CMake run during the SDist" gate, `get_requires_for_build_sdist`, the hatch plugin, and the setuptools shim need no changes (they already key off `sdist.cmake`).

## Changes

- **`settings/skbuild_model.py`** — new `sdist.install-dir`; `sdist.cmake` docstring documents the configure-only vs install-vendor split.
- **`build/wheel.py`** — `_build_wheel_impl` grows a `cmake_install_dir` staging path: it runs the normal configure+build+install (state `"sdist"`), copies the `CMAKE_INSTALL_PREFIX` tree out, and returns *before* any wheel is assembled. Errors clearly if CMake is disabled.
- **`build/sdist.py`** — `sdist.cmake` + `sdist.install-dir` stages the tree into a temp dir kept alive by the existing `ExitStack`, then adds it to the tar under `sdist.install-dir` (reuses `iter_force_include`, sorted for reproducible member order). Warns on the install-dir-without-cmake misconfig.
- **Schema + docs** — regenerated `scikit-build.schema.json`, `configs.md`, `README`; added a "Building a CMake-free wheel from the SDist" section to `docs/configuration/index.md`.
- **Tests** — a `LANGUAGES NONE` header+config test package proving the install tree (incl. a configure-time version-stamped file) is vendored into the SDist, that a wheel from the unpacked SDist is `py3-none-any` with no CMake configure step, that it matches a direct CMake build file-for-file, and that `install-dir` without `cmake` warns and no-ops.

## Open question

- **What to capture**: currently the `CMAKE_INSTALL_PREFIX` (platlib) tree, which covers pybind11/nanobind. Projects that install to other wheel trees (`data`/`headers`/`scripts`) would need a follow-up.

## Verification

- `prek -a --quiet` clean (mypy/ruff)
- `tests/test_sdist_cmake_install.py`, `tests/test_schema.py`, `tests/test_settings_docs.py`, `tests/test_get_requires.py`, `tests/test_force_include.py`, `tests/test_hatchling.py` pass; configure-only path (`sdist.cmake = true`) still passes its existing integration test


henryiii:
GPT 5.5:

The new feature works for the narrow tested case, but it can silently omit supported CMake install destinations and can produce different output for projects with wheel-state-specific configuration.

Full review comments:

- [P2] Copy installs from all wheel staging trees — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/wheel.py:448-449
  When `sdist.install-dir` is used on a project that installs files outside the CMake prefix via the supported `SKBUILD_HEADERS_DIR`, `SKBUILD_SCRIPTS_DIR`, `SKBUILD_DATA_DIR`, or `SKBUILD_METADATA_DIR` cache variables, those files are present in a direct wheel but are dropped here because only `install_dir` (usually platlib/purelib) is copied into the vendored tree. The wheel later built from the sdist can therefore miss headers, scripts, or data that the same CMake install would have produced; the staged wheel trees need to be captured or this unsupported case should be rejected.

- [P2] Preserve wheel-state behavior for vendored installs — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/wheel.py:156-157
  For the `sdist.install-dir` path, this labels the configure/install pass as `sdist`, so `if.state = "wheel"` overrides and CMake checks of `SKBUILD_STATE == "wheel"` are not applied even though the produced install tree is meant to be repackaged as the wheel contents. Projects that customize their wheel install through state-specific overrides will get a different from-sdist wheel than a direct wheel build.

henryiii:

Both review findings are addressed in `33a94e8`.

**[P2] Copy installs from all wheel staging trees** — `sdist.install-dir` now vendors *every* non-empty staged wheel tree, not just the install prefix. Each tree lands under its own subdirectory in the vendored dir (`platlib`/`purelib`, `data`, `headers`, `scripts`, `metadata`), so files installed via `SKBUILD_HEADERS_DIR` / `SKBUILD_SCRIPTS_DIR` / `SKBUILD_DATA_DIR` / `SKBUILD_METADATA_DIR` are preserved. The discard tree (`null`) is skipped. The from-sdist override remaps each tree back with `wheel.force-include.".cmake-install/<tree>" = "${SKBUILD_<TREE>_DIR}"`. The test package now also installs into the headers tree, and the file-for-file match test covers it.

**[P2] Preserve wheel-state behavior** — the configure/install pass now runs as `state = "wheel"` instead of `"sdist"` (its output becomes the wheel contents), so `if.state = "wheel"` overrides and `SKBUILD_STATE == "wheel"` checks apply, matching a direct wheel build. The configure-only `sdist.cmake` path keeps `state = "sdist"`. The test package captures `SKBUILD_STATE` at configure time and asserts it is `"wheel"` (verified to fail without the fix).

Docs and the `sdist.install-dir` reference were updated to show the per-tree mapping.

henryiii:
I think this is getting too complicated for 1.0. Anyway, GPT 5.5:

GPT 5.5:

The new sdist install/repackage flow works for direct hook calls, but the PEP 517 requirement hooks are not updated to match the new behavior. This can cause isolated builds to request unnecessary CMake for repackaged wheels or miss the CMake requirements needed to create the vendored install tree.

Full review comments:

- [P2] Suppress CMake requirements for auto-repackaged sdists — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/wheel.py:163-163
  For an unpacked sdist with `PKG-INFO` and the vendored `sdist.install-dir`, this disables CMake only inside `build_wheel`/`prepare_metadata`. PEP 517 frontends call `get_requires_for_build_wheel` before either hook, and that path still sees the default `wheel.cmake = true`, so isolated/offline builds can fail while trying to install `cmake`/`ninja` even though the later build hook would skip CMake entirely.

- [P2] Include wheel-state CMake requirements for sdist install pass — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/wheel.py:229-233
  When `sdist.install-dir` is set, the sdist build now runs the CMake install pass using `state = "wheel"`, so `if.state = "wheel"` overrides for things like `cmake.version`, `ninja.version`, or generator settings affect the sdist build. However `get_requires_for_build_sdist` still computes requirements only from the `sdist` state, so a frontend can prepare an environment with the wrong/missing CMake or Ninja requirements and then fail during the sdist build.

PR #1400, opened 2026-06-30
fix(editable): refuse to wipe a populated rebuild-dir

henryiii:

## Summary

Follow-up to #1375. As [reported there](https://github.com/scikit-build/scikit-build-core/pull/1375#issuecomment-4847019134), setting `editable.rebuild-dir` to a populated directory (e.g. `python/src` where the `.py` source lives) deleted the user's entire source folder, because the rebuild install tree is unconditionally `shutil.rmtree`'d on each build.

We can't really make rebuild-dir overlap a source tree *work* — the tree must be wiped each build since `cmake --install` only ever adds files, so stale artifacts would otherwise linger. But we can stop destroying source.

## Change

- New `prepare_editable_rebuild_dir` helper drops a `CACHEDIR.TAG` into any tree we create for a user-chosen `rebuild-dir`. The standard signature line opts the tree out of backup tools; the `scikit-build-core` line in it doubles as our ownership marker.
- It also writes a `.gitignore` (`*`) so the compiled artifacts stay out of version control. This mirrors what tools like Cargo do for their `target/` build trees.
- Before wiping a user-chosen `rebuild-dir`, refuse (raise `FileExistsError` with a clear message) when the directory is **non-empty and does not carry our cache-tag** — i.e. something we didn't create.
- Empty / non-existent / previously-managed trees are still wiped and refreshed as before, so reinstalls and import-triggered rebuilds keep working.
- The default `build-dir/install/<targetlib>` tree is fully owned by us, so it stays unguarded (`guard=False`) — no behavior change, no tag/gitignore written, and no risk to existing `editable.rebuild = true` users.

Instead of silently deleting `python/src/mypackage`, the user now gets:

> `editable.rebuild-dir … is not empty and was not created by scikit-build-core; refusing to delete it. Point rebuild-dir at a new or empty directory -- not at a source tree.`

## Tests

- 3 new unit tests: refuses a populated dir (source preserved), refreshes its own tagged tree (and writes `CACHEDIR.TAG` + `.gitignore`), default tree wiped unguarded with no tag/gitignore.
- Existing `test_editable_rebuild_dir` integration test still passes (the new files don't interfere with import resolution).
- Docstring updated; `nox -t gen` regenerated `docs/reference/configs.md`.

### DIFF
--- a/docs/reference/configs.md
+++ b/docs/reference/configs.md
@@ -443,6 +443,10 @@ print(mk_skbuild_docs())
   the same template substitutions as :confval:`build-dir`. This relocates only
   the install tree; :confval:`build-dir` is still required and still hosts the
   CMake build that the rebuild re-runs.
+
+  The tree is wiped and recreated on each build, so it must be a fresh or
+  scikit-build-core-managed directory -- pointing it at a populated directory
+  such as your source tree is refused to avoid deleting those files.
 ```
 
 ```{eval-rst}
--- a/docs/reference/configs.md
+++ b/docs/reference/configs.md
@@ -446,7 +446,9 @@ print(mk_skbuild_docs())
 
   The tree is wiped and recreated on each build, so it must be a fresh or
   scikit-build-core-managed directory -- pointing it at a populated directory
-  such as your source tree is refused to avoid deleting those files.
+  such as your source tree is refused to avoid deleting those files. A managed
+  tree gets a ``CACHEDIR.TAG`` and a ``.gitignore`` so its compiled artifacts
+  stay out of backups and version control.
 ```
 
 ```{eval-rst}
--- a/src/scikit_build_core/build/common_wheel_helpers.py
+++ b/src/scikit_build_core/build/common_wheel_helpers.py
@@ -52,9 +52,17 @@ __all__ = [
     "prepare_wheel_dirs",
 ]
 
-# Dropped into a user-chosen editable.rebuild-dir so a later build knows the tree
-# is ours and is safe to wipe and refresh.
-REBUILD_DIR_MARKER = ".skbuild-editable"
+# A cache-directory tag (https://bford.info/cachedir/) dropped into a user-chosen
+# editable.rebuild-dir. The standard first line opts the tree out of backups; the
+# scikit-build-core line doubles as our ownership marker, so a later build knows
+# the tree is ours and is safe to wipe and refresh.
+CACHEDIR_TAG_NAME = "CACHEDIR.TAG"
+CACHEDIR_TAG_CONTENT = (
+    "Signature: 8a477f597d28d172789f06886806bc55\n"
+    "# This file marks a scikit-build-core editable rebuild-dir install tree.\n"
+    "# It is wiped and refreshed on every build; do not store anything here.\n"
+    "# For information about cache directory tags see https://bford.info/cachedir/\n"
+)
 
 WheelState = Literal[
     "sdist", "wheel", "editable", "metadata_wheel", "metadata_editable"
@@ -142,6 +150,12 @@ def get_editable_rebuild_dir(
     return (build_dir / "install" / targetlib).resolve()
 
 
+def _is_owned_rebuild_dir(targetlib_dir: Path) -> bool:
+    """True if ``targetlib_dir`` carries our cache-tag, i.e. we created it."""
+    tag = targetlib_dir / CACHEDIR_TAG_NAME
+    return tag.is_file() and "scikit-build-core" in tag.read_text(encoding="utf-8")
+
+
 def prepare_editable_rebuild_dir(targetlib_dir: Path, *, guard: bool) -> None:
     """
     Clean and (re)create the persistent install tree for a rebuildable editable.
@@ -154,11 +168,16 @@ def prepare_editable_rebuild_dir(targetlib_dir: Path, *, guard: bool) -> None:
     A user-chosen ``editable.rebuild-dir`` (``guard=True``) may accidentally
     point at a populated directory -- e.g. the package's own source tree. Wiping
     that would delete the user's source (#1135), so refuse unless the directory
-    is empty or carries the marker we drop on a tree we created ourselves.
+    is empty or carries the cache-tag we drop on a tree we created ourselves. A
+    guarded tree also gets a ``.gitignore`` so its compiled artifacts stay out of
+    version control.
     """
     if targetlib_dir.exists():
-        marker = targetlib_dir / REBUILD_DIR_MARKER
-        if guard and not marker.is_file() and any(targetlib_dir.iterdir()):
+        if (
+            guard
+            and not _is_owned_rebuild_dir(targetlib_dir)
+            and any(targetlib_dir.iterdir())
+        ):
             msg = (
                 f"editable.rebuild-dir {targetlib_dir} is not empty and was not "
                 "created by scikit-build-core; refusing to delete it. Point "
@@ -168,7 +187,13 @@ def prepare_editable_rebuild_dir(targetlib_dir: Path, *, guard: bool) -> None:
         shutil.rmtree(targetlib_dir)
     targetlib_dir.mkdir(parents=True)
     if guard:
-        (targetlib_dir / REBUILD_DIR_MARKER).touch()
+        (targetlib_dir / CACHEDIR_TAG_NAME).write_text(
+            CACHEDIR_TAG_CONTENT, encoding="utf-8"
+        )
+        (targetlib_dir / ".gitignore").write_text(
+            "# Created by scikit-build-core; an editable rebuild install tree.\n*\n",
+            encoding="utf-8",
+        )
 
 
 def prepare_wheel_dirs(wheel_root: Path, *, targetlib: TargetLib) -> dict[str, Path]:
--- a/src/scikit_build_core/build/common_wheel_helpers.py
+++ b/src/scikit_build_core/build/common_wheel_helpers.py
@@ -11,6 +11,7 @@ configure/build/install. The wheel-assembly side (WheelWriter vs hatchling
 from __future__ import annotations
 
 import os
+import shutil
 import sysconfig
 from pathlib import Path
 from typing import TYPE_CHECKING, Literal
@@ -47,8 +48,13 @@ __all__ = [
     "get_targetlib",
     "get_wheel_tag",
     "install_wheel",
+    "prepare_editable_rebuild_dir",
     "prepare_wheel_dirs",
 ]
+
+# Dropped into a user-chosen editable.rebuild-dir so a later build knows the tree
+# is ours and is safe to wipe and refresh.
+REBUILD_DIR_MARKER = ".skbuild-editable"
 
 WheelState = Literal[
     "sdist", "wheel", "editable", "metadata_wheel", "metadata_editable"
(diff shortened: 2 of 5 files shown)


WardBrian:
> We can't really make rebuild-dir overlap a source tree work — the tree must be wiped each build since cmake --install only ever adds files, so stale artifacts would otherwise linger. But we can stop destroying source.

I think this is untrue. `build-dir` [will contain an `install-manifest.txt`](https://gitlab.kitware.com/cmake/community/-/wikis/FAQ#can-i-do-make-uninstall-with-cmake) after the `cmake --install` invocation, which lists each file's full install path. Removing just those (or, even better, those that are a sub-tree of `editable.rebuild-dir`) should allow this.

To elaborate more on why I'd like this, it means that my code can just always look in `Path(__file__).parent`, and it will succeed whether this is the editable install or the final installation, without needing importlib

LecrisUT:
> I think this is untrue. `build-dir` will contain an `install-manifest.txt` after the `cmake --install` invocation, which lists each file's full install path

Isn't this generator specific?

But also it's been a while, but doesn't CMake check if the install files are up to date?

> To elaborate more on why I'd like this, it means that my code can just always look in `Path(__file__).parent`, and it will succeed whether this is the editable install or the final installation, without needing importlib

But the question is if you should. For one workflow it works, but what about your general users?

Otherwise it seems you are looking for `editable.mode=inplace` or maybe a derivative of it?

henryiii:
I don't think you should point the rebuild dir at valid folders in your project. Agree, it sounds like you want `editable.mode=inplace`. Though the rebuild doesn't work with that, since we don't intercept anything with inplace (we can't intercept `Path(__file__).parent`!). Also, though, you should not use `Path(__file__).parent`, you should use `importlib.metadata.files()`, you aren't supposed to manipulate file paths directly with Python's import system. There's no guarantee that the filesystem is real (it can be run from a zip file, for example). (this is theoretical, most of the time it's fine, but this is an example where it isn't, and generally it's not correct, only practical).

WardBrian:
>Isn't this generator specific? 

I don't believe so, it appears to be [generated by cmake itself](https://gitlab.kitware.com/cmake/cmake/-/blob/master/Source/cmLocalGenerator.cxx#L785-786)

> But the question is if you should. For one workflow it works, but what about your general users?

I'm not sure what you mean, but placing the compiled code in the same folder as the .py files under `site-packages/` works well for end users. The difficulty arises only in the editable mode, where the `.py` files will live somewhere else.

> Otherwise it seems you are looking for editable.mode=inplace or maybe a derivative of it?

I am hoping I can get the outcome (`libmypackage.so` in `python/src/mypackage` when installed with `-e`) without all the other "caveats" of the inplace mode. In particular, because the python library I'm building is just a part of a larger codebase, "littering the source directory with build artifacts" is a nonstarter)

It's not dramatically worse to have to use importlib for the editable mode only if it's not in scope to support. The additional check before nuking the directory is nice!

henryiii:
What caveats of inplace mode are a problem? Maybe that's fixable.

WardBrian:
The doc lists three ("only one build per source directory, you can’t change to an out-of-source builds without removing the build artifacts, your source directory will be littered with build artifacts"), all of which are things I'd like to avoid personally

---

I realized another downside with using `importlib.resources.files` which is that it will only work if I use platform-dependent filenames. I was previously using [`np.ctypeslib.load_library`](https://numpy.org/doc/2.3/reference/routines.ctypeslib.html#numpy.ctypeslib.load_library) to avoid having to care about .so/.dll/.dylib, but if I'm to use `importlib.resources.files("mypackage") / "libmypackage.so"` I will be forced to re-create that logic myself to get the correct behavior.

LecrisUT:
> I was previously using [`np.ctypeslib.load_library`](https://numpy.org/doc/2.3/reference/routines.ctypeslib.html#numpy.ctypeslib.load_library) to avoid having to care about .so/.dll/.dylib, but if I'm to use `importlib.resources.files("mypackage") / "libmypackage.so"` I will be forced to re-create that logic myself to get the correct behavior.

Fair point. Only recommendation I have for these are glob patterns and not having to do `load_library` in the first place, but it seems you already found good workarounds for those. I don't know if we have it documented, but the `load_library` approach does not work on either pypy or alpine, don't remember which one.

henryiii:
How do you expect to have more than one build per source directory if you put the outputs of the build into the source directory? Two of the three issues basically boil down to "the output is in the source". The other is that you can't do both, which is a bit more of a problem if you also want to run a standard build.

You can get filenames from importlib.resources, from a glob, and then feed that into load_library.

WardBrian:
I didn’t realize globs were an option, that likely solves that. 

You’re correct that I am not expecting to get more than one build for the Python library in this setup. But I would like to be able to build other non-Python parts of the project without those drawbacks, so I guess it really is the fact that both can’t be done at once that’s the main issue

henryiii:

Good — I have what I need. Here's the analysis.

## Short answer

Yes, the `install_manifest.txt` approach WardBrian proposed is technically viable and would work for a class of projects — but it trades a *guaranteed-clean* install tree for a *best-effort* one, and it only *probabilistically* protects source instead of categorically refusing to touch it. That trade-off is why the guard approach was chosen instead.

## How the manifest approach would work

`cmake --install` writes `install_manifest.txt` into the **build dir** (not the install prefix), listing the absolute path of every file it installed. So instead of:

```python
shutil.rmtree(targetlib_dir)   # nuke everything, then reinstall
```

we'd do the classic "make uninstall" recipe: read the *previous* manifest, `os.remove` each listed path, then reinstall. Because we delete by the old manifest *before* the new install overwrites it, renames/removed targets are handled correctly. This surgically removes only our own artifacts and leaves a co-located `foo.py` untouched — exactly the `libmypackage.so` next to source that WardBrian wants.

On the factual sub-debate in the thread: **WardBrian is right and LecrisUT's worry is unfounded** — the manifest is emitted by CMake's `cmLocalGenerator`, not by a specific generator, so it's portable across Ninja/Make/MSVC.

## Why it wasn't the chosen path

The downsides are real and several of them re-introduce the exact problems `rmtree` exists to prevent:

1. **It only *probabilistically* protects source.** The manifest contains whatever CMake installed. The common case (a few compiled libs) is safe. But a `CMakeLists` that does `install(DIRECTORY src/mypackage/ ...)` or `install(FILES foo.py ...)` into the overlap would list hand-written files in the manifest — and we'd delete them. The guard approach categorically refuses to delete anything it didn't create; the manifest approach *trusts* that everything CMake installed is disposable, which is usually but not always true.

2. **Best-effort, not guaranteed-clean.** Manifests list files but never the directories CMake created (empty dirs leak), and the manifest can drift or vanish — if someone cleans `build-dir` but not the overlapping `rebuild-dir`, there's no record, so you'd either leave stale artifacts (the thing we're avoiding) or fall back to wiping (the thing that's dangerous over source).

3. **Multi-config bookkeeping.** Each `cmake --install` overwrites the manifest, so with extra build types installed into the same prefix you'd only have the last config's record. Solvable by accumulating manifests, but more state to get right.

Worth noting: the import-time rebuild in the redirect shim (`_editable_redirect.py:446-456`) *already* doesn't wipe — it just re-runs `cmake --build` + `cmake --install` over the existing tree. So the system already tolerates non-wiped installs on rebuild; the full `rmtree` only happens at the initial wheel build. That actually narrows the gap and makes the manifest approach more plausible than it first appears — but it's also why stale-artifact accumulation isn't catastrophic in practice today.

## The deeper reason

The maintainer position (Henry's comments) is that overlapping build output with source is the anti-pattern, not the goal. It re-creates the inplace-mode caveats (one build per source dir, littered source tree) and leans on `Path(__file__).parent`, which isn't a guaranteed-real filesystem path — the correct API is `importlib.resources.files()` / `importlib.metadata.files()`. So the guard isn't trying to make the overlap *work*; it just stops a foot-gun from deleting source and steers people toward `editable.mode=inplace` or proper resource APIs.

**Bottom line:** the manifest path could have shipped and would satisfy the requesting user's workflow, but it would make scikit-build-core endorse and own a fragile overlap pattern. The chosen guard is the conservative complement: don't enable the anti-pattern, just stop it from destroying data.


---

> Thoughts for a way to make this opt-in?

---

Here's how I'd design an opt-in for it.

## The setting

Add a tri-state knob on the editable rebuild cleanup, rather than a bare boolean. The danger isn't "overlap source" per se — it's "delete by manifest without an ownership marker." So gate the *cleanup strategy*, not the overlap:

```toml
[tool.scikit-build.editable]
rebuild-dir = "python/src/mypackage"
rebuild-dir-cleanup = "manifest"   # "wipe" (default) | "manifest" | "none"
```

- **`wipe`** (default) — today's behavior: `rmtree` the tree, guarded so it refuses a populated unmarked dir. No change for anyone.
- **`manifest`** — opt into the "make uninstall" path: delete only the paths in the previous `install_manifest.txt`, then reinstall. This is what lets the tree overlap source safely-ish.
- **`none`** — never clean; just reinstall over the top (accepts stale-artifact risk; useful for the inplace-overlap crowd who'd rather leak than risk deletion).

A single boolean (`overlap-source = true`) also works, but `manifest`/`none` are genuinely different cleanup semantics with different failure modes, so a string is more honest about what the user is choosing.

## What `manifest` mode does at build time

1. **Don't write `.gitignore = *`.** That's correct for a dedicated build tree, catastrophic over source. In manifest mode either skip it or write a manifest-scoped ignore listing only the installed artifacts (you have the exact paths).
2. **Still drop `CACHEDIR.TAG`?** No — opting out of backups on a source tree is wrong. The ownership marker becomes the *manifest itself*: "I will only ever delete paths this manifest recorded."
3. **Keep the guard, repurposed.** On the very first build there's no prior manifest, so a pre-existing same-named artifact just gets overwritten (fine). But if a manifest path coincides with a git-tracked file, that's the real footgun — so the high-value safety check is: **before deleting a manifest entry, refuse (or warn) if it's tracked by git.** That's a far better source-protection heuristic than "is the dir empty," and it's cheap (`git ls-files --error-unmatch`, or check against the sdist file list you already compute).

## Where it slots in the code

`prepare_editable_rebuild_dir` (`common_wheel_helpers.py:159`) becomes a dispatch on the cleanup mode. The manifest read is straightforward — `install_manifest.txt` lands in `build_dir`, which that function would need passed in (it currently only gets `targetlib_dir`). Multi-config needs the accumulate-manifests fix (each `cmake --install` overwrites it), so you'd snapshot/append per config in `build_install_extra_build_types`.

## The honest caveat to document

Even opt-in, this blesses `Path(__file__).parent` resolution, which Henry's thread argued against. I'd frame the docs as: *"This is for projects that must co-locate compiled output with source in editable mode. It does not clean perfectly (empty dirs leak, manifest can drift) and you are responsible for not installing over hand-written files. Prefer `importlib.resources.files()` and a dedicated rebuild-dir if you can."* Opt-in plus a git-tracked guard plus a loud docstring is enough to ship it without endorsing it as the happy path.

## Smallest viable version

If you want to minimize surface area: skip the tri-state, add just `rebuild-dir-cleanup = "manifest"` as an `experimental`/override-only field, reuse the existing guard machinery, and lean on the git-tracked check as the safety net. That gets WardBrian unblocked with one new field and no change to any default.

Want me to sketch the `prepare_editable_rebuild_dir` dispatch + the manifest-delete helper as a concrete diff?

WardBrian:
Two quick things:
- `.glob` is not actually a method on [Traversable](https://docs.python.org/3/library/importlib.resources.abc.html#importlib.resources.abc.Traversable), so using it with importlib.resources.files would be relying on the implementation detail that the return is a Path
- I'd be surprised if globs were smart enough to actually work with the metapath magic

But I think .iterdir would probably work for this, or I could just write the janky logic to get the dll extension

PR #1403, opened 2026-06-30
feat(editable): expose rebuild() on redirected module loaders

henryiii:

## What

Lets a user trigger an editable rebuild on demand using only the standard library, by calling `rebuild()` on the loader of any redirected import:

```python
import some_package

some_package.__loader__.rebuild()
```

This runs the same `cmake --build`/`--install` cycle that the optional import-time auto-rebuild uses, but under user control — without needing to enable `editable.rebuild`.

## How

The redirecting finder (`resources/_editable_redirect.py`) now wraps **every** spec it returns in a loader exposing a `rebuild()` hook:

- New `_ScikitBuildLoaderWrapper` — a thin delegating wrapper that adds only `rebuild()`. Used for single-location packages, plain modules, and compiled extension modules (the previously-bare cases).
- `_ScikitBuildResourceLoaderWrapper` now extends it, so multi-location packages keep their cross-tree resource reader **and** gain `rebuild()`.
- `_ScikitBuildNamespaceLoader` gains `rebuild()` too.

`find_spec` only needs the heavier resource wrapper when a package has more than one search location; everything else gets the simple wrapper.

### Errors when it can't rebuild

`finder.rebuild()` gained a `required` flag:

- The import-time auto-rebuild calls it leniently → still a silent no-op when there's no persistent build directory.
- The user-facing loader hooks call `rebuild(required=True)` → raises `RuntimeError` pointing at `-Cbuild-dir=build` when no build dir is configured.

cmake build/install failures continue to raise `CalledProcessError` as before.

## Notes

- A rebuild needs a persistent build directory, so install with `-Cbuild-dir=build`.
- Single-location packages were already wrapped (a 1-element search-location list is truthy); switching them to the lean wrapper is behavior-preserving for resources, since their single search location always equals the origin's directory.

## Docs

Added a "Triggering a rebuild manually" subsection under Editable installs in `docs/configuration/index.md`, including the `sys.meta_path` fallback for grabbing the finder directly.

## Tests

- `test_loader_exposes_rebuild` — covers all three loader kinds (single-location package, compiled module, namespace package) and confirms non-rebuild attributes still delegate.
- `test_loader_rebuild_without_build_dir_errors` — the loader call raises while the import-time path stays a silent no-op.

`prek` (ruff/mypy) clean; editable test suites pass.

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -828,11 +828,10 @@ location, so changes there are picked up on import, regardless of the
 
 ### Triggering a rebuild manually
 
-You don't have to enable `editable.rebuild` to rebuild on demand. The
-redirecting finder installs a loader for each redirected module, and that loader
-exposes a `rebuild()` method, so you can run the same
-`cmake --build`/`--install` cycle whenever you like using only the standard
-library:
+You don't have to enable `editable.rebuild` to rebuild on demand for the
+redirect mode. The redirecting finder installs a loader for each redirected
+module, and that loader exposes a `rebuild()` method, so you can run the same
+`cmake --build`/`--install` cycle whenever you like:
 
 ```python
 import some_package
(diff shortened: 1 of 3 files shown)


LecrisUT:
Huh, neat idea. Wonder if we should not make the basic editable file for all modes.

henryiii:
I've always wanted to do this, but didn't have the patience to work out how to get this properly attached to `__loader__` everywhere.

All modes meaning inplace? I think we only have two modes at the moment. Could be a followup.

PR #1411, opened 2026-07-01
feat(editable): expose rebuild() on inplace editable loaders

henryiii:

## What

Extends the on-demand `module.__loader__.rebuild()` hook — added for **redirect** editables in #1403 — to **inplace** editable installs:

```python
import some_package

some_package.__loader__.rebuild()
```

For inplace, `rebuild()` runs `cmake --build` in the source tree (which *is* the CMake build directory for inplace builds). There is no `cmake --install` step, since inplace compiles extensions straight into the source tree.

Also lifts the previous restriction so import-time `editable.rebuild` is now allowed with `editable.mode = "inplace"`.

## How

- **Slim dedicated finder** (`resources/_editable_redirect.py`): new `ScikitBuildInplaceFinder` that does *not* change import resolution — it delegates to `importlib.machinery.PathFinder` against the same search locations the `.pth` adds — and only wraps the resolved loader in the existing `_ScikitBuildLoaderWrapper` to attach `rebuild()`. It returns `None` for imports outside the installed packages and for namespace packages (no concrete loader), so native resolution is untouched. Compiled submodules of a known package get the hook too.
- **Shared rebuild logic**: the redirect finder's `rebuild()` body is extracted into a module-level `_run_editable_rebuild(...)`; passing `install_dir=None` skips the `cmake --install` step (the inplace case), keeping the file-lock/recursion-guard logic in one place.
- **Generation** (`build/_editable.py`): new `editable_inplace(...)` appends an `install_inplace(...)` call (or a PEP 829 `entrypoint()` for `.start` files); `editable_inplace_files(...)` now emits the shim + `.pth`/`.start` mirroring `editable_redirect_files`, instead of just a bare `.pth`. Both build backends (`build/wheel.py`, `hatch/plugin.py`) pass the packages, source dir, and build options.
- **Settings** (`skbuild_read_settings.py`): `editable.mode=inplace` + `editable.rebuild=true` is now valid and needs no separate `build-dir`; `editable.rebuild-dir` stays rejected for inplace (it relocates the install tree, which inplace lacks).

## Notes / behavior change

Inplace editables now ship a `_editable_skbc_<name>.py` shim (+ `.start` on 3.15+) alongside the `.pth`, where before they shipped only the `.pth`. The `.pth` path entries are kept, so native resolution still works; the finder just adds the `rebuild()` hook.

## Tests

- Unit: finder rebuilds build-only (no `--install`), loader exposes `rebuild()` on package + submodule, ignores unknown imports, debounces import-time rebuild once per process, `install_inplace` dedup, and shim generation (legacy `.pth` / PEP 829 `.start` / baked rebuild flag).
- Settings: inplace + `rebuild` allowed; inplace + `rebuild-dir` rejected.
- Integration: `test_inplace_loader_rebuild` installs inplace and confirms `simplest.__loader__.rebuild()` actually runs `cmake --build`; `test_direct_import` asserts the loader exposes `rebuild()` in both modes. Updated `test_pep660_wheel` and `test_hatchling_editable_wheel` for the new inplace file set.

## Docs

Updated the "Triggering a rebuild manually" section and the inplace-mode note in `docs/configuration/index.md`; regenerated `docs/reference/configs.md`.

`prek` (ruff/mypy/blacken-docs/prettier/cog) clean; full non-network test suite passes.

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -908,10 +908,10 @@ etc. Also, to make your binaries importable, you should set
 `$<0:>` for multi-config generator support, like MSVC, so you don't have to set
 all possible `*_<CONFIG>` variations) to make sure they are placed inside your
 source directory inside the Python packages; this will be run from the build
-directory, rather than installed. This will also not support automatic rebuilds.
-The build directory setting will be ignored if you use this and perform an
-editable install. You can detect this mode by checking for an in-place build and
-checking `SKBUILD` being set.
+directory, rather than installed. The build directory setting will be ignored if
+you use this and perform an editable install (the source directory doubles as
+the build directory). You can detect this mode by checking for an in-place build
+and checking `SKBUILD` being set.
 
 With all the caveats, this is very logically simple (one directory) and a near
 identical replacement for `python setup.py build_ext --inplace`. Some third
@@ -920,6 +920,9 @@ install a `.pth` file that points at your source package(s) and do an inplace
 CMake build.
 
 On the command line, you can pass `-Ceditable.mode=inplace` to enable this mode.
+Inplace installs support both automatic (`editable.rebuild`) and manual rebuilds
+(see below); since the source directory doubles as the build directory, no
+separate `build-dir` is needed.
 
 :::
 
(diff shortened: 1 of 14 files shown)


henryiii:

The patch adds inplace editable rebuild hooks, but misses supported single-module package entries, leaving the advertised hook and auto-rebuild behavior inactive for those projects.

Review comment:

- [P2] Handle module-file entries when computing inplace finder names — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/_editable.py:227-227
  When `wheel.packages` points at a single module file such as `hello.py` (supported by the package mapping code), this records `known_packages` as `"hello.py"`, but `ScikitBuildInplaceFinder.find_spec()` compares it to `fullname.partition(".")[0]` (for `import hello`, `"hello"`). In inplace editables that module then falls through to the normal loader, so `editable.rebuild=true` never triggers and `hello.__loader__.rebuild()` is not exposed for module-only projects.

PR #1415, opened 2026-07-01
perf: adopt lazy imports via flake8-lazy

henryiii:

## What

Adds the [flake8-lazy](https://github.com/henryiii/flake8-lazy) pre-commit hook (`--apply=set`) and runs it once, declaring deferrable imports in `__lazy_modules__` across the codebase. flake8-lazy finds top-level imports only used inside functions and records them so they can be deferred. The declarations are backward compatible on all Python versions and honored natively as lazy imports on **3.15+ (PEP 810)**, cutting import-time overhead.

- `.pre-commit-config.yaml`: new hook, pinned to `v0.8.3`, placed after ruff so declarations stay correct over time.
- 159 `.py` files gained a `__lazy_modules__` declaration (`src/scikit_build_core/_vendor` excluded via the existing top-level `exclude`). No import statements were moved or removed; output is ruff-format compatible and idempotent (a second run is a no-op, no `LZY4xx` diagnostics).

## Why — measured startup impact

Python **3.15.0b2**, editable install, `hyperfine --warmup 10`, toggling the change via `git stash`:

| CLI invocation | Before (eager) | After (lazy) | Δ |
|---|---|---|---|
| `scikit-build` (info / dispatch) | 56.5 ± 1.6 ms | **37.1 ± 1.6 ms** | **−34%** |
| `scikit-build builder` | 96.1 ± 2.6 ms | **90.3 ± 2.8 ms** | −6% |
| `import scikit_build_core.build` (backend) | 14.1 ± 0.9 ms | 13.7 ± 0.9 ms | ~noise |

Modules imported on `import scikit_build_core.__main__`: **100 → 52**.

The `scikit-build` CLI wins most: `main()` builds one argparse tree over all four subcommands, which previously dragged in the whole dependency chain just to dispatch. The build backend is intentionally unaffected — its deferred imports live in `wheel.py`/`sdist.py`, which a real build still executes.

## Verification

- Import smoke check ✓ (all key modules import on 3.15).
- Confirmed lazy behavior: after `import scikit_build_core.build.wheel`, `scikit_build_core.cmake` and `shutil` are absent from `sys.modules` on 3.15.
- Tests ✓ — `pytest -m "not compile and not network and not integration and not isolated and not virtualenv and not setuptools"`: **810 passed, 3 skipped** (env-only skips).

## Notes

- The hook is self-contained (`language: python`), so it is **not** added to project dependencies.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,6 +31,16 @@ repos:
         args: ["--fix"]
       - id: ruff-format
 
+  - repo: https://github.com/henryiii/flake8-lazy
+    rev: e50a00f1c8ed0b55d642dbbec274de848d4c2018 # frozen: v0.8.4
+    hooks:
+      - id: flake8-lazy
+        # Only the shipped package benefits from lazy imports; docs, examples,
+        # tests, and the noxfile don't need it.
+        files: ^src/
+        # --strict-typing guards the relative-import form so it passes mypy.
+        args: [--apply=set, --strict-typing]
+
   - repo: https://github.com/pre-commit/pygrep-hooks
     rev: 3a6eb0fadf60b3cccfd80bad9dbb6fae7e47b316 # frozen: v1.10.0
     hooks:
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -35,10 +35,7 @@ repos:
     rev: e50a00f1c8ed0b55d642dbbec274de848d4c2018 # frozen: v0.8.4
     hooks:
       - id: flake8-lazy
-        # Only the shipped package benefits from lazy imports; docs, examples,
-        # tests, and the noxfile don't need it.
         files: ^src/
-        # --strict-typing guards the relative-import form so it passes mypy.
         args: [--apply=set, --strict-typing]
 
   - repo: https://github.com/pre-commit/pygrep-hooks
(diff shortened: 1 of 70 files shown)


henryiii:

Marked as **draft**: this depends on henryiii/flake8-lazy#87 (mypy-safe `__spec__.parent` for multi-level relative imports). The hook is temporarily pinned to that PR's head commit; it will be repinned to a tagged release before this is marked ready.

Fixes since the first push:
- Excluded `tests/packages/` from the hook and reverted fixture `__lazy_modules__` declarations — flake8-lazy had lazified fixture packages, which broke editable rebuild-on-import on Python 3.15 (`import simplest` no longer triggered the rebuild).
- `__main__.py` files switched to absolute imports (their `__spec__.parent` is `str | None` under strict mypy).
- Regular modules migrated to the guarded `(__spec__.parent or "")` form via flake8-lazy#87.

Verified locally on 3.15.0b3: editable/CLI/init tests pass, mypy passes, hook application is idempotent.

henryiii:

flake8-lazy [#87](https://github.com/henryiii/flake8-lazy/pull/87) is merged and released as **v0.8.4**. The hook is now pinned to `v0.8.4` (frozen SHA) with `--strict-typing`, which emits the guarded `(__spec__.parent or "")` form that passes strict mypy. Re-running the hook produces no diff and mypy is clean locally on 3.15.0b3. Marking ready for review.

Issue #1417, opened 2026-07-01
Fable pre-release review of scikit-build-core


henryiii:
:robot: _AI text below_ :robot:

Review complete. I audited all 132 commits in v0.12.2..HEAD across eight subsystems with parallel review agents, then personally re-verified the highest-severity claims by reproduction. The full report (31 findings with file:line references, plus everything that verified clean) is in the attached file. The headline: **seven issues I'd consider release blockers**, most on brand-new 1.0 features, all confirmed real:

1. **Editable installs of namespace packages are broken** (reproduced myself): the `.pth` points one level too deep, so `import myns.mypkg` fails while `import mypkg` wrongly succeeds. Since #1385 auto-discovers namespace packages, any dotted-layout project silently gets a broken editable install. `build/wheel.py:549-551` + `_editable.py`.
2. **sdist crashes on dangling symlinks** (reproduced myself): the new `resolve-symlinks = "all"` default makes `tar.add` stat every path, so a broken symlink anywhere in the tree raises a raw `FileNotFoundError` out of the PEP 517 hook — a tree that built fine on 0.12.2. `build/sdist.py:141`.
3. **Hatch plugin dies on hatchling's reserved `enable-by-default` key**: the documented on-demand-build pattern exits 7 on every build. One-line fix in `hatch/plugin.py:78-81`, and the plugin just got de-experimentalized.
4. **CC/CXX stripping breaks compiler wrappers** (regression from #1331): `CC="ccache gcc"` becomes `CC=ccache`, failing CMake compiler identification for pyenv/self-built interpreters. Strip only `-…` flag tokens instead. `builder/generator.py:152-160`.
5. **Entry-point config providers' `except TypeError` fallback** double-executes providers and swallows real errors; needs `inspect.signature` dispatch. `settings/_load_entrypoint_config.py:55-58`.
6. **Entry-point config is validated as if it were static pyproject config**: a distro provider setting `cmake.toolchain-file` — the feature's target use case — exits 7 with a misleading error, and providers touching version-gated fields break every project pinning an older `minimum-version`.
7. **`dynamic_wheel` is documented as a supported 0.3-spec hook but never called anywhere** — wire it up or remove it before the API freezes at 1.0.

[...]

PR #1434, opened 2026-07-02
feat(sdist): "external" and "classic" resolve-symlinks modes; "none" stores directory symlinks

henryiii:

Fixes item 2 of #1417 (see [the findings comment](https://github.com/scikit-build/scikit-build-core/issues/1417#issuecomment-4860351655)).

`sdist.resolve-symlinks = "none"` documented "store symlinks as-is", but the walk followed directory symlinks unconditionally, so a `link -> dir` had its contents archived as duplicated regular files with no link member. This PR makes the option's modes match their contracts:

- **`"all"`** (default, unchanged): resolve every symlink, copying its target's contents.
- **`"external"`** (new): store symlinks that stay inside the project as-is (files *and* directories); resolve symlinks whose immediate target points outside the project (absolute target, Windows drive, or a relative target that lexically escapes the root), since those would dangle once the sdist is extracted. Checking only the immediate target keeps chains consistent link-by-link.
- **`"none"`** (fixed): store every symlink as-is, including directory symlinks (stored as link members, walk pruned).
- **`"classic"`** (new): the previous de-facto behavior — file symlinks as-is, directory symlinks followed with contents duplicated. `minimum-version < 1.0` now maps here instead of `"none"` (same observable behavior as 0.x). Named to match `sdist.inclusion-mode = "classic"`.

Symlinks that can't be resolved are stored as link members in **every** mode, with a warning when they were supposed to be resolved:
- dangling symlinks (extends the #1421 fallback to `"external"` mode's dereference path);
- directory symlink loops — previously silently dropped on POSIX but archived on Windows (where `os.walk` classifies the loop as a file), so storing the link uniformly also fixes platform-divergent sdist contents. `add_path_to_tar` refuses to dereference a symlink resolving to a directory (only pruned loops reach it), which would otherwise recurse forever under `"all"`.

Implementation: `each_unignored_file()` gains a `resolve_symlinks` parameter (default `"all"`, so the wheel-packages caller is unchanged) that yields internal directory symlinks — and pruned loop symlinks — as members instead of descending, plus a `symlink_escapes()` helper; `build_sdist` keeps tar-level `dereference` only for `"all"` and adds escaping file symlinks via their resolved target in `"external"` mode. Schema and config reference regenerated.

Tests: 5-way mode parametrization over internal/external file and dir symlinks plus a loop symlink, asserting each member's tar type and preserved link targets; dangling-symlink tests for `"all"` and `"external"`. Each fix had a failing regression test first. Also verified a round trip manually: a `"none"` sdist extracts with working relative symlinks, and an `"external"` sdist ships the outside-the-tree file dereferenced.

### DIFF
--- a/docs/reference/configs.md
+++ b/docs/reference/configs.md
@@ -795,6 +795,10 @@ print(mk_skbuild_docs())
   * "classic": Store file symlinks as-is, but follow directory symlinks,
     copying their contents (scikit-build-core 0.x behavior).
 
+  A symlink that can't be resolved (dangling, or a directory symlink loop)
+  is stored as a symlink in every mode, with a warning if it was supposed to
+  be resolved.
+
   If you don't set this, it will be "all" unless you set the minimum version
   below 1.0, in which case it will be "classic" to preserve backward
   compatibility.
--- a/docs/reference/configs.md
+++ b/docs/reference/configs.md
@@ -778,7 +778,7 @@ print(mk_skbuild_docs())
 ```{eval-rst}
 .. confval:: sdist.resolve-symlinks
 
-  :Type: ``"all" | "none"``
+  :Type: ``"all" | "external" | "none" | "classic"``
   :Default: "all"
   :Config-settings: ``sdist.resolve-symlinks`` or ``skbuild.sdist.resolve-symlinks``
   :Environment variable: ``SKBUILD_SDIST_RESOLVE_SYMLINKS``
@@ -788,10 +788,16 @@ print(mk_skbuild_docs())
   The modes are:
 
   * "all": Resolve every symlink, copying its target's contents.
-  * "none": Store symlinks as-is.
+  * "external": Resolve only symlinks that point outside the project (those
+    would be broken once the SDist is extracted); store symlinks that stay
+    inside the project as-is.
+  * "none": Store every symlink as-is, including directory symlinks.
+  * "classic": Store file symlinks as-is, but follow directory symlinks,
+    copying their contents (scikit-build-core 0.x behavior).
 
   If you don't set this, it will be "all" unless you set the minimum version
-  below 1.0, in which case it will be "none" to preserve backward compatibility.
+  below 1.0, in which case it will be "classic" to preserve backward
+  compatibility.
 
   .. versionadded:: 1.0
 ```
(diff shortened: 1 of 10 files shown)


henryiii:
GPT 5.5 in Codex:

The patch adds useful symlink modes, but it introduces a wheel-build regression for circular directory symlinks and can still create broken sdists in external mode for symlinks nested under followed external directories.

Full review comments:

- [P2] Avoid yielding loop directory symlinks to wheel copying — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/_file_processor.py:165-165
  When a package source contains a circular directory symlink, this new `yield dirpath` also affects `packages_to_file_mapping()` because it calls `each_unignored_file()` with the default `resolve_symlinks="all"`. Non-editable wheel builds then try `shutil.copy2()` on that yielded directory symlink and fail with `IsADirectoryError`; before this change the loop was pruned without being yielded. Please keep this sdist-specific or otherwise avoid returning directory symlinks to the wheel package-copy path.

- [P2] Resolve symlinks under followed external directories — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/_file_processor.py:55-56
  In `external` mode, after following an external directory symlink, child symlinks are still classified lexically using the archive path. For example, with `ext_dir -> ../outside/dir` and `outside/dir/link -> ../secret.txt`, `joined` becomes `ext_dir/../secret.txt`, so the link is treated as internal and archived as a symlink even though its real target is outside the project and is not copied into the sdist, leaving a broken link after extraction. This needs to account for the resolved parent when walking through an external symlinked directory.

henryiii:

Both review findings are addressed:

- **Loop directory symlinks reaching wheel copying**: `each_unignored_file()` now only yields a pruned loop symlink when the new `yield_loop_symlinks` flag is set, which only the SDist walk passes. Wheel package copying gets the previous prune-silently behavior back, so `shutil.copy2()` never sees a directory symlink. Regression tests at the `packages_to_file_mapping()` and walk levels. (86ecf53)

- **Symlinks nested under followed external directories**: `symlink_escapes()` now judges containment against the nearest followed symlink ancestor rather than the project root, since only that subtree mirrors the real directory the link resolves against. So `ext_dir/link -> ../secret.txt` is dereferenced (both the file-symlink path in `sdist.py` and the directory-symlink path in the walk go through this check), while links staying inside the followed directory are still stored as symlinks. The sdist symlink test matrix now covers nested internal/escaping file and directory symlinks in all four modes. (29fa796)

PR #1441, opened 2026-07-02
feat(editable): serve force-include entries live via the redirect

henryiii:

Previously, `wheel.force-include` entries were copied into a redirect-mode editable wheel, so edits to those files silently did nothing until reinstall. Now the import redirect serves them live from their sources where it can:

- **Importable modules** — always, including renamed destinations (the finder loads the source path as the destination module).
- **Data files** — when they keep their filename and land inside a package (the source directory joins the package `__path__`, so `importlib.resources` sees the live file).

A redirected entry displaces a package file or CMake output at the same destination (the finder checks installed files before sources, so the stale staged file is removed), and force-included directory members remain subject to `wheel.exclude`.

Everything the redirect cannot express keeps the previous copy behavior: other wheel trees (`${SKBUILD_SCRIPTS_DIR}` etc.), renamed or top-level data files, and all of inplace mode.

No compatibility gate: `force-include` is unreleased (new in 1.0).

Beyond the new tests, verified end-to-end by building an editable wheel with a force-included renamed module and data file, importing through the shim from an extracted site-packages, editing the sources, and observing the changes in a fresh process without reinstalling.

### DIFF
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -442,6 +442,15 @@ destination. A force-included _directory_ stays subject to that exclude, so a
 bulk tree copy can still be trimmed by an exclude pattern (e.g. force-include a
 directory and exclude `**/*.bzl` to drop the Bazel files from it).
 
+In an editable install using redirect mode (the default), platlib entries are
+served live from their sources through the import redirect instead of being
+copied, so edits to a force-included file take effect without reinstalling. This
+covers importable modules (even renamed ones) and data files that keep their
+filename and land inside a package. Anything the redirect cannot express — other
+wheel trees like `${SKBUILD_SCRIPTS_DIR}`, renamed data files, or top-level data
+files — is still copied at install time, so those snapshots only refresh on
+reinstall.
+
 #### Building a wheel from an SDist
 
 A common pattern vendors an external (`../`) source into the SDist and then
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -446,10 +446,11 @@ In an editable install using redirect mode (the default), platlib entries are
 served live from their sources through the import redirect instead of being
 copied, so edits to a force-included file take effect without reinstalling. This
 covers importable modules (even renamed ones) and data files that keep their
-filename and land inside a package. Anything the redirect cannot express — other
-wheel trees like `${SKBUILD_SCRIPTS_DIR}`, renamed data files, or top-level data
-files — is still copied at install time, so those snapshots only refresh on
-reinstall.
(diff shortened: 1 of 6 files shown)


henryiii:
GPT 5.5:

The patch changes editable redirect behavior for force-included nested data files in a way that makes package-root resource lookups diverge from normal wheels and previous editable wheels.

Review comment:

- [P2] Keep nested package data copied in editable wheels — /Users/henryfs/git/scikit-build/scikit-build-core/src/scikit_build_core/build/_pathutil.py:376-376
  When a same-name data file is force-included below a package subdirectory, e.g. `wheel.force-include = {"assets" = "pkg/assets"}`, this condition marks `pkg/assets/a.txt` as redirectable and stops baking it into the editable wheel. The redirect only registers the source directory as `pkg.assets`, so common resource access through the package root such as `importlib.resources.files(pkg).joinpath("assets/a.txt")` no longer finds the file, while the normal wheel and the previous editable copy did. Please either keep these nested data files copied or register the ancestor package paths so package-root resource lookups still work.

henryiii:

Addressed in 19f6970, plus the Windows/cygwin CI failures.

**Review finding (nested data files):** fixed as suggested — the data-file redirect is now restricted to files directly inside a top-level package (`len(rel.parts) == 2`); anything deeper (e.g. `pkg/assets/a.txt` from a directory force-include) is baked into the wheel again, preserving the wheel layout. Modules still redirect at any depth, since imports resolve by name rather than by `__path__` traversal from an ancestor.

One correction to the finding's premise, verified empirically against `main`: `importlib.resources.files("pkg").joinpath("assets/a.txt")` did **not** work in the previous editable copy behavior either — the redirect finder only registers install-tree files under their immediate parent (`pkg.assets`), and `pkg.__path__` never contains a real `assets/` directory. `files("pkg.assets").joinpath("a.txt")` works before and after. So the package-root traversal gap is a pre-existing finder limitation that equally affects CMake-installed nested data files, and would be a separate fix (registering install-tree ancestor directories); this change restores the physical-copy semantics rather than fixing that. I did not register source-tree ancestors as the alternative suggested, since for a source like `assets/` mapped to `pkg/assets` that would put the project root on `pkg.__path__`.

**CI (Windows 3.8, cygwin 3.9):** test assertion encoding, not behavior — the shim embeds `repr()`-escaped paths built from `os.getcwd()`, so the expected values need `repr(str(Path.cwd()/...))`: raw `str(path)` never matches doubled backslashes on Windows, and on cygwin `tmp_path` carries 8.3 short names (`RUNNER~1`) while `getcwd()` returns long names.

PR #1443, opened 2026-07-02
test: cover uv as an install frontend via a workspace

### LINKED ISSUE #1328: tests: cover uv as an install frontend (editable + build isolation)

Follow-up from review discussion on #1325 (https://github.com/scikit-build/scikit-build-core/pull/1325#discussion_r3401469060).

The integration suite currently exercises pip exclusively as the install frontend (`VEnv.install`). #1325 added `VEnv.aux_install`, which uses uv (when on PATH) for tooling installs, but deliberately keeps every install that is itself under test on pip so coverage does not depend on what happens to be on PATH.

As @LecrisUT pointed out, the frontends differ in ways that matter to a build backend, so at least one editable-mode and one build-isolation test should run against **both** pip and uv:

- pip's build isolation is a sys.path-based "fake venv", while uv creates a real venv — backend code that inspects the running environment can behave differently.
- Editable install handling (PEP 660 wheel consumption, `.pth` placement) also differs between the two.

Proposed approach:

- Parametrize a small, representative set over the frontend — e.g. one `test_editable.py` case (redirect + inplace) and one isolated-build case (`pip install .` vs `uv pip install .` / `uv build`).
- Skip the uv rows when `shutil.which("uv")` is `None` (matching the `aux_install` fallback).
- For the wheelhouse, uv needs explicit `--no-index --find-links` arguments rather than the `PIP_NO_INDEX`/`PIP_FIND_LINKS` env vars `VEnv.run` sets; `VEnv.aux_install` already shows the pattern.

henryiii:

Closes #1328.

The integration suite installed the package-under-test only via pip. This adds a uv-only test that drives `uv sync` on a uv workspace with two scikit-build-core members — a compiled C extension and a pure `wheel.cmake = false` consumer depending on it via `{ workspace = true }`. That exercises uv's real install path (build isolation in a real venv, PEP 660 editable members) rather than parameterizing existing pip tests.

Asserts the extension imports through the consumer, both per-package editable redirect shims coexist in one environment, and the member's Python sources are served live from the workspace checkout. Runs offline against the test wheelhouse with a per-test uv cache; skipped when uv is not on PATH (it is present in the main CI test matrix via setup-uv).

### DIFF
--- a/tests/test_uv_workspace.py
+++ b/tests/test_uv_workspace.py
@@ -1,0 +1,123 @@
+from __future__ import annotations
+
+import os
+import shutil
+import subprocess
+import sys
+import textwrap
+from pathlib import Path
+
+import pytest
+
+DIR = Path(__file__).parent.resolve()
+
+UV = shutil.which("uv")
+
+pytestmark = [
+    pytest.mark.compile,
+    pytest.mark.configure,
+    pytest.mark.integration,
+    pytest.mark.network,
+    pytest.mark.virtualenv,
+    pytest.mark.skipif(UV is None, reason="uv is not on PATH"),
+]
+
+
+def test_uv_workspace_editable_member(tmp_path: Path, pep518_wheelhouse: Path):
+    """A scikit-build-core package as a uv workspace member.
+
+    ``uv sync`` installs workspace members editable through the standard PEP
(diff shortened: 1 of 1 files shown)


henryiii:
Instead of making a parametrized test for pip and uv, made this a uv workspace test, to test both things. :)

PR #1445, opened 2026-07-02
feat: add {name} placeholder for template-formatted settings

henryiii:

## Summary

In uv/hatch workspaces, users could set one build directory for all workspace members via `SKBUILD_BUILD_DIR` or a workspace-root `[tool.uv] config-settings` entry — those apply to every member uv builds, so all members currently collide in one CMake build dir. This adds a `{name}` placeholder (the normalized project name) to `build-dir` / `editable.rebuild-dir` / `build.requires` templating, so a single shared setting is safe, e.g.:

```
SKBUILD_BUILD_DIR=/path/to/cache/{name}/{cache_tag}
```

`{name}` is normalized like the wheel/sdist filename (PEP 503 canonicalized, then `-` replaced by `_`: lowercase, `_`-separated), matching the convention already used in `build/sdist.py` and `build/_wheelfile.py`.

`{name}` is available anywhere `build-dir` and `editable.rebuild-dir` are resolved (both the PEP 517 wheel/sdist backend and the hatchling plugin), since project metadata is loaded before those steps. It is **not** available when resolving `build.requires` (`GetRequires.dynamic_metadata`), since that runs before metadata is read; using `{name}` there raises a `KeyError`.

### DIFF
--- a/docs/configuration/formatted.md
+++ b/docs/configuration/formatted.md
@@ -21,6 +21,9 @@ visibility
 .. autoattribute:: scikit_build_core.format.PyprojectFormatter.cache_tag
    :no-index:
 
+.. autoattribute:: scikit_build_core.format.PyprojectFormatter.name
+   :no-index:
+
 .. autoattribute:: scikit_build_core.format.PyprojectFormatter.root
    :no-index:
 
--- a/docs/configuration/index.md
+++ b/docs/configuration/index.md
@@ -1020,7 +1020,9 @@ speedups.
 ```
 
 There are several values you can access through Python's formatting syntax. See
-[](./formatted.md).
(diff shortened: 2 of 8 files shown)


henryiii:

Closing — we decided we don't need a `{name}` placeholder for now. Workspace guidance is covered by #1444, and per-member build dirs work with `{root}`-relative or per-member configuration.

henryiii:

Reopening — closed by mistake; this stands on its own (the workspace test PR just won't depend on it).

henryiii:
Fable was a little too quick there to assume my intent. :)

henryiii:
(And Sonnet 5 is terrible, had Fable launch a Sonnet 5 subagent for this and it made a mistake that had to be fixed)

PR #1448, opened 2026-07-03
docs: dedupe, simplify, and restructure the documentation

henryiii:

A three-part cleanup of the docs (net −305 lines while adding four pages), from a full review pass:

- **In-place dedup and polish**: repeated prose consolidated (getting-started CMake boilerplate was in 6 of 8 tabs, dynamic.md explained the legacy metadata table five times, build.md had twin SDist/Wheel hook lists), platform sections and more settings converted to tabs/`conf-tabs`, condition and `py-api` walls of prose converted to tables, asides moved into admonitions. Also fixes outdated `cmake.minimum-version` advice in the migration guide.
- **Page ownership**: topics that were told in full on 2–3 pages (wheel repair, Finding Python, `SKBUILD_*_DIR` trees, parallel builds) now have one owning page; other mentions shrink to a sentence + crosslink, with reciprocal see-also links added.
- **Structural moves**: new pages `guide/ctypes.md`, `guide/debugging.md` (coverage + IDE IntelliSense merged around their shared root cause), `configuration/editable.md`, and `configuration/variants.md` — the FAQ shrinks to true Q&A stubs and the configuration guide no longer links into the FAQ for feature docs.

Old external deep links to moved sections are not redirected (discussed; intentional). Docs build warning-free from a clean tree.

### DIFF
--- a/docs/_static/custom.css
+++ b/docs/_static/custom.css
@@ -1,0 +1,5 @@
+/* Let the sidebar scroll past the last entry, so the bottom menu items can sit
+   higher than the viewport edge. */
+.sidebar-tree {
+  padding-bottom: 8rem;
+}
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -182,6 +182,7 @@ html_static_path = ["_static"]
 # it (and the box) from furo's theme variables. tippy.css must come last so its
 # overrides win. sphinx-tippy already loads tippy.js itself from this CDN.
 html_css_files = [
+    "custom.css",
     "https://unpkg.com/tippy.js@6/themes/light-border.css",
     "tippy.css",
 ]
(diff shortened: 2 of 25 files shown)


henryiii:
Notes: (will edit as I review):

- Getting started: The tree has been deduped, I liked it better as tabs
- Getting started: Numpy note feels a bit out of place, 2.0+ is common these days
- Some of these pages might make sense with one more nesting level, might be worth checking the theme to see if that's supported.
- The FAQ still has sections that are now their own page, just deleting the FAQ entry would make sense.
- Tabs were added to the configuration page, but inconsistently, probably should either have all, none, or only ones that are likely to be passed in
- Dynamic-metadata: side note, we support 0.4+ now, not 0.3
- Dynamic-metadata: The broken tabs before are now not tabs at all. Would prefer modern/classic tabs.

PR #1439, opened 2026-07-02
chore(skill): add test-downstream A/B regression-check skill

henryiii:
:robot: _AI text below_ :robot:

## What

Adds a `test-downstream` agent skill that answers one question: does the current
checkout regress a downstream project versus the last release? It builds the
project twice per mode — baseline tag vs working tree, in both a wheel build and
an editable install — through `nox -s downstream` and reports a parity table
plus wheel-content and warning diffs. The primary signal is **parity**: a mode
that built on the baseline but fails on the current checkout is a regression.

Both sides run in their own detached git worktree, so each gets a private
`.nox/downstream/` venv and clone tree. The current worktree faithfully replays
the dirty working tree (tracked diff vs HEAD + untracked, non-ignored files)
onto HEAD, so uncommitted changes are still what's tested. That isolation lets
multiple A/B runs execute concurrently on one checkout without clobbering each
other's `.nox`.

## Projects exercised (current `main` vs `v0.12.2`)

Ran the skill against **20** downstream projects. Recorded here so future runs
can pick up where this left off (the curated list is `docs/data/projects.toml`).

**✅ parity (16):** iminuit, spglib, rapidfuzz, gemmi, gdstk, boost-histogram,
h3, Levenshtein, nanobind, osqp, pyzmq, tetgen, pygram11, deflate, pyhmmer,
highspy

**🔴 regression (2):**
- **manifold3d** — `build/_pathutil.py` (from #1395) now hard-raises when a
  `wheel.packages` entry's source doesn't exist on disk. manifold's `manifold3d`
  module is produced by CMake install, so v0.12.2 skipped it silently, current
  aborts. Fix in #1440.
- **coreforecast** — `settings/skbuild_read_settings.py` (from #1399) turns a
  renamed field set without `minimum-version` into an error. coreforecast uses
  `cmake.verbose` (renamed to `build.verbose`) without pinning `minimum-version`,
  so a WARNING became an aborting ERROR. Intended behavior of #1399, but it
  breaks any project using an old field name without a pinned minimum-version.

**⚪ upstream-only failure, not us (2):** astyle (malformed author email rejected
by its setuptools root), symusic (C++20 structured-bindings error in its vendored
`zpp_bits.h`) — both fail identically on baseline and current.

**Not yet run** (rest of `docs/data/projects.toml`): adios2, afdko, antspyx,
awkward-cpp, celerite2, clang-format, cmake, ducc0, faiss-cpu, freud-analysis,
[...]

PR #1449, opened 2026-07-03
feat: harden downstream A/B harness for real-world projects

henryiii:
Followup to #1439.

Running the downstream A/B skill across many projects surfaced three harness limitations (not backend bugs) that stopped projects from ever reaching scikit-build-core. This fixes them and adds two inputs.

**Clone-dir sanitization.** The clone dir was `"_".join(project.split("/"))`, which for a URL leaves the `://` colon in the path (`https:__github.com_...`). That embedded colon made `pyproject_hooks` mis-split a project's `backend-path`, so e.g. ninja died with `Cannot find module 'backend'` before the backend ran. Extracted a `downstream_dir_name` helper that maps runs of non-`[A-Za-z0-9._-]` to `_` and drops a trailing `.git`, so the name is colon-free and stable across re-runs. Unit-tested (regression: no colon for the ninja-style URL).

**Subdir pyproject read.** The session ran `load_toml("pyproject.toml")` at the clone root but only chdir'd into `--subdir` afterward, so every subdir project (lightgbm, awkward-cpp, ompl, kiss-icp, pyarrow) failed to read its build pyproject on both baseline and current. Reordered so `--prepare` runs at the root, then the subdir chdir happens, then the pyproject read + requires strip/install run in the subdir.

**New inputs.**
- `--requires` (repeatable): extra packages installed into the no-isolation env after the project's own requires — for a dynamic-metadata provider a project needs but doesn't declare (ninja needs `hatch-fancy-pypi-readme`).
- `--prepare`: a repo-local prep command run once at the clone root before configure (awkward-cpp needs the parent repo's `nox -s prepare` to generate kernel headers).

Both are threaded through `downstream_ab.py` into the nox posargs (not the git-clone passthrough) and documented in SKILL.md, including the rule that `--prepare` must not mutate the host — system-package installs are container-only.

Also wired the two known cases into `docs/data/projects.toml` (`requires` for ninja, `prepare` for awkward-cpp); the docs cog only reads `pypi`/`github`/`path`, so the extra keys are ignored by generation.

### DIFF
--- a/.agents/skills/test-downstream/SKILL.md
+++ b/.agents/skills/test-downstream/SKILL.md
@@ -71,11 +71,15 @@ Each entry has:
 pypi = "iminuit"            # what the user is likely to say
 github = "scikit-hep/iminuit"   # -> https://github.com/scikit-hep/iminuit
 path = "awkward-cpp/pyproject.toml"   # optional -> --subdir awkward-cpp
+requires = ["hatch-fancy-pypi-readme"]  # optional -> --requires (repeatable)
+prepare = "nox -s prepare"    # optional -> --prepare (repo-local prep command)
 ```
 
 - The `project` argument to nox is a git URL: `https://github.com/<github>`.
 - If `path` is present, pass `--subdir <dirname of path>` (strip the trailing
   `/pyproject.toml`).
+- If `requires` is present, forward each as `--requires <pkg>`.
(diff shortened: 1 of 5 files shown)


henryiii:
Let's not add this, at least now. I think we can fix awkward-cpp, and the extra might go away with the new dynamic-metadata.

PR #1455, opened 2026-07-04
Bump pyproject-metadata bundling info

LecrisUT:
Followup to #1454

Also dropping the obsoletes as no longer needed

### DIFF
--- a/.distro/python-scikit-build-core.spec
+++ b/.distro/python-scikit-build-core.spec
@@ -38,7 +38,7 @@ Requires:       cmake
 Requires:       ninja-build
 BuildArch:      noarch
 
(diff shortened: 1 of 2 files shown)


henryiii:

Updated `nox -s vendor_pyproject_metadata` to also sync the bundled version into the distro spec. After vendoring, the session reads `__version__` from the freshly-vendored `pyproject_metadata/__init__.py` and rewrites the `Provides: bundled(python3dist(pyproject-metadata))` line in `.distro/python-scikit-build-core.spec`, so it can't drift out of date again. Also corrected the stale `0.12.0` → `0.12.1` to match what's actually vendored.

Reading the version from the vendored source rather than the session's `version` arg keeps it correct even when that arg is a branch, a `v`-prefixed tag, or a commit sha.

PR #1483, opened 2026-07-07
fix(editable): don't truncate shared namespaces in redirect mode

### LINKED ISSUE #1482: Problem with editable installs
We noticed an issue with editable installs during the upgrade to 1.0.0. Below is an analysis generated by Opus 4.8 via Claude Code. I hope that's okay.

Some context: [`mqt-ddsim`](https://github.com/munich-quantum-toolkit/ddsim) and [`mqt-qmap`](https://github.com/munich-quantum-toolkit/qmap) are C++ libraries with Python bindings (via `nanobind`). Both depend on [`mqt-core`](https://github.com/munich-quantum-toolkit/core/). After upgrading to 1.0.0, `mqt-ddsim` can no longer find `mqt-core` (that is, in a clean `mqt-ddsim` environment, `uv run python -c "import mqt.core"` fails), while `mqt-qmap` still can.

```
Now I have the complete picture. Here's exactly what's happening:

---

Root Cause: Missing install-tree path in known_directories for the mqt namespace

scikit-build-core 1.0.0 generates a ScikitBuildRedirectingFinder that is inserted at sys.meta_path[0]. When Python does import mqt, this finder intercepts the import and creates a namespace spec whose __path__ is entirely controlled by the known_directories argument in the generated install() call at the bottom of _editable_skbc_mqt_*.py.

ddsim's generated install() — known_directories for mqt:
'mqt': ['/Users/daniel/mqt/ddsim/python/mqt']   # source tree only

qmap's generated install() — known_directories for mqt:
'mqt': ['/Users/daniel/mqt/qmap/python/mqt', 'mqt']   # source tree + install tree

The second 'mqt' in qmap's list is a relative path that resolves to {site-packages}/mqt. With that in __path__, Python can find mqt.core there. Without it (ddsim), the finder hijacks the namespace and limits it to the source directory — mqt.core becomes invisible.

Why ddsim is missing it:

scikit-build-core decides to include the install-tree path for a namespace package by walking up from registered install-tree subdirectories. qmap's install tree has mqt/qmap/na/ (a directory containing .pyi stubs), which gets registered as mqt.qmap.na. That triggers walking up through mqt/qmap → mqt, so all three get the install-tree path.

ddsim's install tree has only mqt/ddsim/pyddsim.abi3.so — no subdirectory. mqt.ddsim gets the install-tree path mqt/ddsim, but there's no subdirectory trigger to also walk up to mqt. So mqt never gets 'mqt' (the install-tree entry), and the finder truncates the namespace.

Why ~=0.12 fixes it:

scikit-build-core 0.12.x generated editable installs that only added the source tree to sys.path via a plain .pth entry — no meta path finder at all. Python's standard namespace discovery merged site-packages/mqt with python/mqt naturally. The 1.0.0 redirecting finder is what breaks the namespace.

Summary: it's a bug in scikit-build-core 1.0.0 — when a leaf-level package has an install-tree path (mqt.ddsim → mqt/ddsim), the parent namespace (mqt) should also receive the install-tree parent path (mqt), but it only does so if there is a grandchild install-tree subdirectory that triggers the upward walk.
```

henryiii:

Fixes #1482.

The redirecting finder synthesized shared (PEP 420) namespace packages with a `__path__` built only from the tracked directories, truncating the namespace: portions from sibling distributions (e.g. mqt-core installed normally next to an editable mqt-ddsim) disappeared, so `import mqt.core` failed. qmap only survived by accident — a stray `__init__.pyi` under `mqt/qmap/na/` triggered the ancestor walk.

Two layers, both fixed:

- **Generation** (`collect_search_locations`): the namespace-ancestor walk (#1417) only ran for `__init__` entries, so an install tree holding just a leaf module (`mqt/ddsim/pyddsim.abi3.so`) never gave the ancestor (`mqt`) the install-tree path. A second pass now lets loose files extend *already-registered* ancestors; unregistered containing directories (a `wheel.install-dir` prefix) still are not synthesized, preserving #1427.
- **Runtime** (redirect finder): a synthesized namespace spec now merges the portions native `PathFinder` resolution finds on `sys.path` (only when native resolution also sees a namespace), so siblings on any path entry stay importable — matching the pre-1.0 plain-`.pth` behavior. This also covers siblings outside the editable's own site-packages (user-site, other `.pth` entries).

Regression tests were written first and reproduced the report, including an end-to-end venv test where a normally-installed sibling shares the namespace with the editable install and failed with the reported `ModuleNotFoundError`.

### DIFF
--- a/src/scikit_build_core/build/_editable.py
+++ b/src/scikit_build_core/build/_editable.py
@@ -433,6 +433,26 @@ def collect_search_locations(
                     ancestor_dir = str(Path(ancestor_dir).parent)
                     directories.setdefault(ancestor, set()).add(ancestor_dir)
 
+    # Second pass: extend already-registered ancestors with each loose file's
+    # matching parent directories. An install tree holding only leaf files
+    # under a namespace (e.g. ``myns/mypkg/_ext.so``, no install-tree
+    # ``__init__``) otherwise leaves the namespace ancestor (``myns``) without
+    # the install-tree path, and the finder truncates the shared namespace to
+    # the source tree, hiding sibling distributions in site-packages (#1482).
+    # Only ancestors registered above are extended, so an untracked containing
+    # directory (a ``wheel.install-dir`` prefix) is still not synthesized.
+    registered = frozenset(directories)
+    for module, directory, is_init in entries:
+        if is_init:
+            continue
+        ancestor = module.rpartition(".")[0]
+        ancestor_dir = directory
+        while "." in ancestor:
+            ancestor = ancestor.rpartition(".")[0]
+            ancestor_dir = str(Path(ancestor_dir).parent)
+            if ancestor in registered:
+                directories[ancestor].add(ancestor_dir)
+
     return (
         {pkg: sorted(dirs) for pkg, dirs in directories.items()},
         sorted(packages),
--- a/src/scikit_build_core/resources/_editable_redirect.py
+++ b/src/scikit_build_core/resources/_editable_redirect.py
@@ -504,6 +504,18 @@ class ScikitBuildRedirectingFinder(importlib.abc.MetaPathFinder):
         # A tracked package directory without an importable __init__ is a
         # namespace package.
         if submodule_search_locations is not None:
+            # A PEP 420 namespace can be shared with other distributions, so
+            # merge in the portions native resolution would find on sys.path
+            # (e.g. a sibling distribution installed normally into
+            # site-packages); synthesizing from the tracked directories alone
+            # would truncate the namespace (#1482).
+            from importlib.machinery import PathFinder
+
+            native = PathFinder.find_spec(fullname, path)  # type: ignore[arg-type]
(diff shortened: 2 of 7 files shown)


LecrisUT:
Can we get a functional test with 2 packages probably with either same build backend or different ones contributing to the same namespace and see how that behaves. That seems to be the core setup in #1482. I think when we changed to our custom Loader we opened that can of worms

henryiii:

Added a functional no-compile test in aebabb8: two pure packages share the `myns` namespace — `myns.mypkg` (scikit-build-core, `wheel.cmake = false`, installed editable) and a new `myns.otherpkg` fixture (hatchling, installed normally into site-packages). The test pip-installs both into one venv and imports both, parametrized over `redirect` and `inplace` editable modes. Verified the redirect case fails on main with `ModuleNotFoundError: No module named 'myns.otherpkg'` and passes with this fix (inplace passes either way).

PR #1484, opened 2026-07-07
fix: wheel timestamp clamp overflows on 32-bit time_t platforms

### LINKED ISSUE #1481: test_wheel_timestamp_clamps_epoch_beyond_zip_range test fails on 32bit
```
self = WheelWriter(metadata=StandardMetadata(name='something', version=<Version('1.2.3')>, description=None, license=None, li...build_tag=''), metadata_dir=None, variant_label='', variant_dist_info_contents=None, reproducible=False, _zipfile=None)
mtime = None
    def timestamp(
        self, mtime: float | None = None
    ) -> tuple[int, int, int, int, int, int]:
        if self.reproducible:
            # Honor SOURCE_DATE_EPOCH, else a fixed epoch (ignore per-file mtime).
            timestamp = get_reproducible_epoch()
        else:
            raw = os.environ.get("SOURCE_DATE_EPOCH")
            timestamp = (
                parse_source_date_epoch(raw)
                if raw is not None
                else int(mtime or time.time())
            )
        # The ZIP file format only supports timestamps from 1980-01-01 to
        # 2107-12-31 23:59:59 (inclusive); clamp rather than let zipfile crash.
        timestamp = min(max(timestamp, MIN_TIMESTAMP), MAX_TIMESTAMP)
>       return time.gmtime(timestamp)[0:6]
               ^^^^^^^^^^^^^^^^^^^^^^
E       OverflowError: timestamp out of range for platform time_t
../BUILDROOT/usr/lib/python3.15/site-packages/scikit_build_core/build/_wheelfile.py:146: OverflowError
```
If it's possible to pytest skip this easily let's try to do so

henryiii:

Fixes #1481.

`WheelWriter.timestamp()` clamps out-of-range `SOURCE_DATE_EPOCH` values to the ZIP format's limits, but the upper bound (2107-12-31) is past what a 32-bit `time_t` can represent, so `time.gmtime` raised `OverflowError` on 32-bit platforms — in the production code path, not just the test. The conversion now uses pure `datetime` arithmetic, and the test's expected values are hardcoded (they previously called `time.gmtime(MAX_TIMESTAMP)` too).

Also adds a 32-bit i686 Linux CI job. `actions/setup-python` has no 32-bit Linux builds and GHA `container:` can't run i386 images (the runner injects a 64-bit node binary), so the job runs the suite via `docker run --platform linux/386` with apt-provided cmake/ninja/g++ and no numpy (no i686 wheels; those tests skip). Both the failure and the fix were verified in a local `i386/python:3.13-slim` container.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -239,6 +239,34 @@ jobs:
           verbose: true
           token: 6d9cc0e0-158a-41ee-b8f4-0318d3595ac2
 
+  checks-32bit:
+    name: 🐍 3.13 • 32-bit Linux (i686)
+    needs: changes
+    if: fromJSON(needs.changes.outputs.run-tests)
+    runs-on: ubuntu-latest
+    timeout-minutes: 40
+    steps:
+      - uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
+        with:
+          persist-credentials: false
+
+      # actions/setup-python has no 32-bit Linux builds, and `container:` can't
+      # run i386 images (the runner injects a 64-bit node), so use docker
+      # directly. i386 runs natively on the amd64 host. No numpy: no i686
+      # wheels; numpy-dependent tests skip. linux32 makes uname report i686.
+      - name: Test package
+        run: |
+          docker run -i --rm --platform linux/386 -v "$PWD:/io" -w /io \
+            i386/python:3.13-slim linux32 bash -euxs <<'EOF'
+          apt-get update
+          apt-get install -y --no-install-recommends cmake make ninja-build g++ git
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -239,6 +239,34 @@ jobs:
           verbose: true
           token: 6d9cc0e0-158a-41ee-b8f4-0318d3595ac2
 
+  checks-32bit:
```
LecrisUT, line 242: I don't think the tests are that much useful, the problems are such infrequent that we can just handle them post release with downstream backports

henryiii: It's faster than some of the surrounding tests (6min vs. up to 9 min), so I don't think it hurts.

PR #1489, opened 2026-07-07
feat: package-declared config-settings

henryiii:
Design choices:

* User config-settings require a `.` (namespaced)
* Explicit strings required (keeps schema simple and avoids TOML dict collisions)

Implements the design discussed in #1468 (targeting **1.1**, not 1.0): packages can declare their own config-settings in a `tool.scikit-build.config-setting` table, giving users a documented, package-level interface instead of raw CMake defines.

```toml
[tool.scikit-build.config-setting."zmq.prefix"]
help = "Prefix to search for libzmq"
env = "ZMQ_PREFIX"          # env > -C > default

[tool.scikit-build.config-setting."zmq.libzmq"]
default = "system"

[tool.scikit-build.cmake.define]
ZMQ_PREFIX = {config-setting = "zmq.prefix"}   # forwarded to CMake; dropped when unset

if.config-setting."zmq.libzmq" = "bundled"     # matches the resolved value
cmake.define.ZMQ_LIBZMQ = "ON"
```

Notes:

- The declaration table is popped from the raw TOML like `overrides` (not a settings field, not an override target); values resolve before override processing so `if.config-setting` sees them.
- Declared keys are whitelisted in `ConfSource` (new `extra_keys`) and get "did you mean" suggestions; `skbuild.`-prefixed custom keys stay rejected, and unknown-key strict-config behavior is otherwise unchanged.
- Declarations are pyproject-only (rejected in overrides, extra settings, and entry-point config) and gated on `minimum-version >= 1.1`.
- Types: `str` and `bool` (strtobool semantics); repeated `-C` values error. A `choices` key for enumerated values was cut to keep the initial surface minimal (a code comment notes how to re-add it).
- The CMake binding is the `cmake.define = {config-setting = "..."}` reference form only; an earlier inline `cmake =` declaration alias was dropped as redundant.
- Also fixes a pre-existing crash where a malformed non-table `cmake` value raised `AttributeError` (in the `force_auto_cmake` lookup) instead of the normal conversion error; the new resolution helpers are guarded the same way.

Beyond CI: verified end-to-end through the real `build_wheel` hook that the env var and `-C` spellings land in CMake identically and the override fires.

### DIFF
--- a/docs/api/scikit_build_core.settings.rst
+++ b/docs/api/scikit_build_core.settings.rst
@@ -25,6 +25,14 @@ scikit\_build\_core.settings.auto\_requires module
    :show-inheritance:
    :undoc-members:
 
+scikit\_build\_core.settings.config\_settings module
+----------------------------------------------------
+
+.. automodule:: scikit_build_core.settings.config_settings
+   :members:
+   :show-inheritance:
+   :undoc-members:
+
 scikit\_build\_core.settings.documentation module
 -------------------------------------------------
 
--- a/docs/configuration/config_settings.md
+++ b/docs/configuration/config_settings.md
@@ -21,7 +21,6 @@ config-settings:
 [tool.scikit-build.config-setting."zmq.prefix"]
 help = "Prefix to search for libzmq"
 env = "ZMQ_PREFIX"
-cmake = "ZMQ_PREFIX"
 
 [tool.scikit-build.config-setting."zmq.libzmq"]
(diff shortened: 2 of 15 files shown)


henryiii:
Dropping a review point from the simplification work.

Unifying {config-setting = ...} with the {env = ...} annotation resolver (altitude's top finding) — a real architectural point, but it means threading resolved values into the generic source-conversion layer; too large a redesign for this pass

PR #1491, opened 2026-07-09
chore(fedora): Backport downstream changes

LecrisUT:
- Missed adding the setuptools and hatchling subpackages
- f2py-cmake and cython-cmake have been packaged. Would take about 1 week for the stable branches to get it https://bodhi.fedoraproject.org/updates/FEDORA-2026-853945cfd5

Will move it out of draft when f2py-cmake and cython-cmake land on stable branches also

### DIFF
--- a/.distro/plans/examples.fmf
+++ b/.distro/plans/examples.fmf
@@ -1,14 +1,5 @@
 summary: Documentation examples
 discover+:
   filter: "tag: examples"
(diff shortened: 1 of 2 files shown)


henryiii:
This is passing now, does that mean it's ready? :)

henryiii (approved):

PR #1493, opened 2026-07-09
fix(editable): resolve redirected modules through sys.path_hooks

### LINKED ISSUE #1492: Feature Request: New option to improve integration with third-party PEP 302 (sys.path_hooks) import hooks
(Disclaimer: I am far from an expert in Python build systems. This is just after quite the journey of trying to make [Fandango](https://github.com/fandango-fuzzer/fandango) work with `scikit-build-core` without losing support for [Beartype](https://github.com/beartype/beartype). See https://github.com/beartype/beartype/issues/556.)

In editable-install mode, `scikit-build-core` adds `ScikitBuildRedirectingFinder` and `ScikitBuildInplaceFinder` to `sys.meta_path`. These finders resolve and load a package's modules directly, without going through the standard `sys.path_hooks` mechanism defined by [PEP 302](https://peps.python.org/pep-0302).

This breaks interoperability with import hooks that are specifically designed to be PEP 302-compliant and avoid `sys.meta_path` contention such as beartype's `beartype.claw`, which hooks via `sys.path_hooks` precisely so it can coexist with other tools' import machinery (e.g. pytest's). Because (to the best of my knowledge) `scikit-build-core`'s finders short-circuit `find_spec()` without consulting `sys.path_hooks`, modules loaded through them are silently never seen by `beartype.claw` (or any other well-behaved `sys.path_hooks`-based hook) — no error, just silent loss of instrumentation.

Currently, we rely on the following hack to get this to work with Fandango: Install the binary as a top-level module (with a `DESTINATION .` in `CMakeLists.txt`), then dynamically remove the two `scikit-build-core` finders at the top of the main `__init__.py`:

```python
import sys

from beartype.claw import beartype_this_package

def _is_fandango_editable_redirect_finder(finder: object) -> bool:
    """True for this package's scikit-build-core redirect editable finder only."""
    if type(finder).__name__ != "ScikitBuildRedirectingFinder":
        return False
    return "fandango" in getattr(finder, "pkgs", ())

sys.meta_path[:] = [
    finder
    for finder in sys.meta_path
    if not _is_fandango_editable_redirect_finder(finder)
]

beartype_this_package()
```

This seems less than optimal. I would love an alternative mode of `scikit-build-core` that works inline with what is expected by PEP 302 and is therefore compatible with Beartype without these hacks. Refer to https://github.com/beartype/beartype/issues/556 for additional details on the problem.

henryiii:

Fixes #1492.

In redirect editable mode, `ScikitBuildRedirectingFinder` built specs with `spec_from_file_location`, hard-wiring the stock `SourceFileLoader` and never consulting `sys.path_hooks`. PEP 302 instrumenting import hooks — e.g. `beartype.claw`, which registers a path hook whose loader rewrites modules at load time (see beartype/beartype#556) — were silently bypassed.

`_make_spec` now resolves the loader through `importlib.machinery.PathFinder` against the directory holding the mapped file, so registered path hooks are honored. The redirect mapping stays authoritative: the spec is accepted only if it resolves to the exact mapped file (compared case-insensitively), packages get their merged source+build-tree `__path__` reapplied, and anything the path machinery can't reproduce (mismatched filename, `__init__.pyc`, missing file) falls back to the previous behavior. Loader wrapping for `rebuild()` and the multi-location resource reader are unchanged.

The inplace finder already delegated to `PathFinder` and was unaffected; the new integration test covers both modes with real beartype (added to the test wheelhouse), and a unit test covers the same via a stub instrumenting path hook with no new dependency. Both were verified to fail before the fix.

No new option is needed, and downstream workarounds that strip the finder from `sys.meta_path` can be dropped.

### DIFF
--- a/src/scikit_build_core/resources/_editable_redirect.py
+++ b/src/scikit_build_core/resources/_editable_redirect.py
@@ -525,11 +525,35 @@ class ScikitBuildRedirectingFinder(importlib.abc.MetaPathFinder):
         submodule_search_locations: list[str] | None,
     ) -> importlib.machinery.ModuleSpec | None:
         is_pkg = origin.endswith(("__init__.py", "__init__.pyc"))
-        spec = importlib.util.spec_from_file_location(
-            fullname,
-            origin,
-            submodule_search_locations=submodule_search_locations if is_pkg else None,
-        )
+        # Resolve the loader through the standard sys.path_hooks machinery (PEP
+        # 302) so instrumenting path hooks (e.g. beartype.claw) see redirected
+        # modules (#1492). Searching only the directory holding our chosen file
+        # keeps the redirect mapping authoritative; the spec is accepted only if
+        # it resolves to that same file.
+        search = os.path.dirname(origin)
+        if is_pkg:
+            search = os.path.dirname(search)
+        spec = importlib.machinery.PathFinder.find_spec(fullname, [search])
+        if (
+            spec is not None
+            and spec.loader is not None
+            and spec.origin is not None
+            and os.path.normcase(spec.origin) == os.path.normcase(origin)
+        ):
+            spec.submodule_search_locations = (
+                submodule_search_locations if is_pkg else None
+            )
+        else:
+            # The mapping cannot be reproduced by the path machinery (e.g. the
+            # file's name does not match the module name), so build the spec
+            # directly with the stock loader.
+            spec = importlib.util.spec_from_file_location(
+                fullname,
+                origin,
+                submodule_search_locations=submodule_search_locations
+                if is_pkg
+                else None,
+            )
         # Wrap the loader so it exposes a rebuild() hook (reachable as
         # module.__loader__.rebuild()). Packages with more than one search
         # location (e.g. a source tree and a CMake install tree) additionally get
--- a/tests/test_editable.py
+++ b/tests/test_editable.py
@@ -309,6 +309,13 @@ def test_editable_with_beartype_claw(isolated, isolate, editable):
     Path("src/simplest/typed.py").write_text("def f(x: int) -> int:\n    return x\n")
 
     isolated.install("beartype")
+    # beartype may not support the newest Python yet (e.g. 0.22.9 fails to
+    # import on 3.15); skip rather than fail, and un-skip automatically once
+    # a supporting release is out.
+    try:
+        isolated.execute("import beartype")
+    except SystemExit:
+        pytest.skip("beartype cannot be imported on this Python")
     isolated.install(
         "-v",
         *isolate.flags,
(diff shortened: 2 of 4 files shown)


LecrisUT:
If this is the fix, it looks like meson-python would have the same issue (also not calling `find_spec` before creating their own finder)
https://github.com/mesonbuild/meson-python/blob/47ea86f0cefa3a5285ae804ff5980b3a178bec43/mesonpy/_editable.py#L255
Shall I forward this issue or would you do that?

henryiii:
Sure, you can.

henryiii:
This morning's release of build added warnings, fixing them in #1494. I want to get that build release yanked and renumbered as 1.6 or 2.0, but we should move off of this deprecated function anyway.

henryiii:
I checked `force-includes`. Standard ones work, renamed source doesn't (it doesn't participate in a filename based finder), and it protects from a leaking nearby name.

Short answer: renamed force-includes fall back to the old (stock-loader) path via the origin-identity guard, so they keep working exactly as before — they just don't get path-hook instrumentation. Un-renamed ones go through path hooks like everything else. Details:

In a redirect-mode editable, platlib force-includes the redirect can serve live are *not* baked into the wheel — `_force_include_into_wheel` (`build/wheel.py:188-200`) drops them into the redirect mapping as `module name → source file`. And `editable_redirectable` (`build/_pathutil.py:378-379`) explicitly permits renames for importable files: `{"src/foo_impl.py" = "pkg/foo.py"}` produces a mapping entry whose on-disk filename (`foo_impl.py`) doesn't match the import name (`pkg.foo`).

That's precisely the case the new `_make_spec` was designed around:

- **Name matches** (the common force-include, and all CMake-installed/package files): `PathFinder.find_spec("pkg.foo", [dir])` finds the same file, the `normcase(spec.origin) == normcase(origin)` check passes, and instrumenting path hooks like beartype.claw see the module. Strictly better than before.
- **Renamed source**: PathFinder looks for `foo.py` in the directory holding `foo_impl.py` and finds nothing → spec is `None` → fallback to `spec_from_file_location`, the pre-fix behavior. The module imports fine; beartype just can't instrument it — which is inherent, since `FileFinder`-based hooks locate modules by filename, and was equally true before the fix.
- **The hazard case the guard exists for**: if the source directory happens to *also* contain a real `foo.py` next to the renamed `foo_impl.py`, PathFinder would return the wrong file. The origin-identity check rejects that spec and falls back, so the force-include mapping stays authoritative and you never silently import the neighbor.

The unit test `test_redirect_falls_back_when_path_hooks_miss` in `tests/test_editable_redirect.py` covers exactly this rename shape (a module mapped to `actual_name.py`).

Force-includes that aren't served live — metadata-tree entries, renamed or nested data files — get baked copies and never touch the finder, so the fix doesn't interact with them at all.

PR #1499, opened 2026-07-09
fix(setuptools): thread config_settings to the build command

henryiii:

The setuptools plugin read settings inconsistently between two code paths. The metadata/validation hooks (`prepare_metadata_for_build_*`, `_validate_editable_settings`) passed PEP 517 `config_settings` to `SettingsReader`, but the `build_cmake` command reloaded settings via `_load_settings` **without** them. Because `build_cmake` runs inside setuptools' `setup()` (invoked by setuptools' `build_wheel`/`build_editable`), it never naturally receives the `config_settings` dict, so they were silently dropped for the actual CMake build.

Concretely, `pip install -e . -C editable.mode=inplace` against a project that does not set `editable.mode` in `pyproject.toml` **passed** the editable metadata hook but then **failed** inside `build_cmake` with `setuptools editable installs require editable.mode = 'inplace'`. Env vars (`SKBUILD_*`) already worked; only `config_settings` were lost.

## Change

- Stash the in-flight `config_settings` in a `contextvars.ContextVar` (`build_cmake.set_config_settings`), scoped via a context manager so nothing leaks between calls.
- All `build_meta` hooks now set the stash before delegating to setuptools: `build_wheel`, `build_editable`, `build_sdist`, `prepare_metadata_for_build_wheel`, and `prepare_metadata_for_build_editable` (`build_wheel`/`build_sdist`/`prepare_metadata_for_build_wheel` are now thin wrappers instead of re-exports).
- `_load_settings` reads the stash, so the build half sees the same config as the metadata half.

This makes `-C` / `--config-settings` work consistently across the setuptools editable and wheel paths, matching the existing env-var behavior.

### DIFF
--- a/docs/plugins/setuptools.md
+++ b/docs/plugins/setuptools.md
@@ -91,8 +91,13 @@ possible. If you don't use that, you get more reasonable modern defaults.
 ## Configuration
 
 All other configuration is available as normal `tool.scikit-build` in
-`pyproject.toml` or environment variables as applicable. Config-settings is
-_not_ supported, as setuptools has very poor support for config-settings.
+`pyproject.toml` or environment variables as applicable. Config-settings (`-C`)
+is supported when building through the PEP 517 backend
+(`scikit_build_core.setuptools.build_meta`), e.g.
+`pip install . -C cmake.build-type=Debug`. It is _not_ available when driving
+`setup.py` directly or via the `wrapper.setup` shim, since setuptools' command
+machinery never receives the PEP 517 `config_settings`; use `SKBUILD_*`
+environment variables there instead.
 
 For classic scikit-build compatibility, two environment variables are honored,
 but only when using the `scikit_build_core.setuptools.wrapper.setup` shim (they
--- a/docs/plugins/setuptools.md
+++ b/docs/plugins/setuptools.md
@@ -94,10 +94,8 @@ All other configuration is available as normal `tool.scikit-build` in
 `pyproject.toml` or environment variables as applicable. Config-settings (`-C`)
 is supported when building through the PEP 517 backend
 (`scikit_build_core.setuptools.build_meta`), e.g.
-`pip install . -C cmake.build-type=Debug`. It is _not_ available when driving
-`setup.py` directly or via the `wrapper.setup` shim, since setuptools' command
-machinery never receives the PEP 517 `config_settings`; use `SKBUILD_*`
-environment variables there instead.
+`pip install . -C cmake.build-type=Debug`. It is _not_ available if you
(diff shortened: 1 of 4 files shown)


henryiii:
Since I've not used ContextVars much, I asked for a MWE:

<details><summary>:robot: MWE :robot:</summary>

```python
"""MWE: how the ContextVar isolates config_settings across two concurrent builds.

Mirrors the setuptools plugin: a build_meta "hook" stashes config_settings, then
calls into a build_cmake "command" that has no way to receive it as an argument
and must read it out of the shared channel.

We force the two builds to interleave (set A, set B, then both read) so the
difference between a ContextVar and a plain global is visible, not luck.
"""

import contextlib
import contextvars
import threading

# Two barriers to choreograph the race deterministically:
#   after_set  — both threads have stashed their settings
#   after_read — both threads have read (so we can compare)
after_set = threading.Barrier(2)

results: dict[str, str] = {}


# ---- ContextVar version (what the PR does) --------------------------------
_CONFIG = contextvars.ContextVar("config_settings", default=None)


@contextlib.contextmanager
def set_config_settings(cfg):
    token = _CONFIG.set(cfg)
    try:
        yield
    finally:
        _CONFIG.reset(token)


def build_cmake_cv(name):
    # The "command": no argument, reads the ambient value.
    after_set.wait()  # ensure the *other* thread has already set its own value
    cfg = _CONFIG.get()
    results[name] = cfg["mode"]


def hook_cv(name, cfg):
    with set_config_settings(cfg):
        build_cmake_cv(name)


# ---- Plain-global version (the alternative) -------------------------------
_GLOBAL = {"cfg": None}


def build_cmake_global(name):
    after_set.wait()
    results[name] = _GLOBAL["cfg"]["mode"]


def hook_global(name, cfg):
    _GLOBAL["cfg"] = cfg
    try:
        build_cmake_global(name)
    finally:
        _GLOBAL["cfg"] = None


def run(hook, label):
    global after_set
    after_set = threading.Barrier(2)
    results.clear()
    a = threading.Thread(target=hook, args=("A", {"mode": "Debug"}))
    b = threading.Thread(target=hook, args=("B", {"mode": "Release"}))
    a.start(), b.start()
    a.join(), b.join()
    ok = results == {"A": "Debug", "B": "Release"}
    print(f"{label:12} -> {results}  {'OK' if ok else 'CROSS-TALK'}")


run(hook_cv, "contextvar")
run(hook_global, "global")
```

</details> 

I don't think it would ever matter in practice, setuptools isn't going to run this in threads. But it's more technically correct than a global, so I'm inclined to keep it.

Also, there's no overlap between our options and setuptools, so not stripping them out currently.

PR #1500, opened 2026-07-09
feat(setuptools): default to inplace editables in the wrapper

henryiii:

setuptools' native editable mechanism is in-place building, and that is how scikit-build (classic) always behaved, so the wrapper (`scikit_build_core.setuptools.wrapper.setup`) no longer requires `editable.mode = "inplace"` for PEP 660 editable installs — it logs an informational line and proceeds. The plain plugin keeps requiring the explicit opt-in, since editable installs write into the source tree.

Since wrapper use is only known once `setup.py` runs, the early PEP 517 validation in `build_meta` defers the `editable.mode` check to the `build_cmake` command, where the wrapper compat flag is available.


henryiii:
I'm going to leave this explicit for now.

PR #1535, opened 2026-08-14
fix: double CMake discovery timeout on native Apple Silicon

### LINKED ISSUE #1530: Native arm64 CMake can exceed discovery timeout in fresh rattler-build prefixes on macOS
### Summary

On native Apple Silicon, scikit-build-core can reject a working conda-forge CMake executable because its first `cmake -E capabilities` invocation in a freshly extracted rattler-build prefix exceeds the fixed five-second discovery timeout. The same executable responds in under one second immediately afterward.

This is similar to #1127 / #1131, but the CMake binary is native arm64 rather than x86_64 under Rosetta.

### Environment

- macOS 26.6.1 on arm64
- Python 3.12.13
- scikit-build-core 1.0.3
- CMake 3.31.8 from conda-forge (`osx-arm64`, `h54ad630_0`)
- rattler-build 0.72.2 with its default strict environment isolation

`file cmake` reports `Mach-O 64-bit executable arm64`.

### Observed behavior

A no-build-isolation pip build reaches CMake discovery and fails after slightly more than five seconds:

```text
2026-08-11 09:32:52,333 - scikit_build_core - WARNING - cmake should not be in build-system.requires - scikit-build-core will inject it as needed
2026-08-11 09:32:57,639 - scikit_build_core - WARNING - Accessing CMake timed out, ignoring
scikit_build_core.errors.CMakeNotFoundError: Could not find CMake with version >=3.15
```

Immediately after that failure, the exact executable succeeds:

```text
$ /usr/bin/time -p cmake -E capabilities
real 0.66
$ python -c "... get_cmake_program(Path(shutil.which(\"cmake\"))) ..."
Program(... version=<Version(\"3.31.8\")>)
elapsed 0.0189
```

I reproduced the fresh-prefix timeout three times. Running `cmake --version` once before invoking pip avoids it, which suggests first-execution verification of the newly extracted ad-hoc-signed binary on macOS.

Setting `CI=1` on the outer rattler-build command does not help because strict build isolation does not forward it into the PEP 517 process. That also means the existing 20-second CI heuristic may not be visible to scikit-build-core when rattler-build is called by CI.

### Question

Would it make sense to give native macOS CMake discovery a longer timeout as well, or to expose a scikit-build-core-specific timeout setting? I can prepare a focused patch and regression test if one of those directions is preferred.

DSeaStar:
Fixes #1530.

A freshly extracted native arm64 CMake can take ~5s on the first `cmake -E capabilities` call, so the 5s base timeout dropped a working CMake. Raising `BASE_TIMEOUT` itself would double CI/Windows/Rosetta timeouts, so this only doubles the native Apple Silicon path (5s → 10s). Rosetta stays at 3×.

## Test plan
- [x] `pytest tests/test_program_search.py`

### DIFF
--- a/src/scikit_build_core/program_search.py
+++ b/src/scikit_build_core/program_search.py
@@ -96,7 +96,9 @@ def compute_timeout(executable: Path) -> int:
         return BASE_TIMEOUT * 2
 
     if sys.platform == "darwin" and platform.machine() == "arm64":
-        return BASE_TIMEOUT * 3 if _macos_binary_is_x86(executable) else BASE_TIMEOUT * 2
+        return (
+            BASE_TIMEOUT * 3 if _macos_binary_is_x86(executable) else BASE_TIMEOUT * 2
+        )
 
     return BASE_TIMEOUT
 
--- a/src/scikit_build_core/program_search.py
+++ b/src/scikit_build_core/program_search.py
@@ -50,7 +50,7 @@ def __dir__() -> list[str]:
     return __all__
 
 
-BASE_TIMEOUT = 10
+BASE_TIMEOUT = 5
 
 
 @functools.lru_cache(None)
@@ -84,6 +84,7 @@ def compute_timeout(executable: Path) -> int:
     * Is on CI: quadruples it
     * Is windows: doubles it
     * Runs with Rosetta: triples it
+    * Native Apple Silicon: doubles it
 
     These do not stack.
     """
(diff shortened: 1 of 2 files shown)


src/scikit_build_core/program_search.py
```diff
@@ -50,7 +50,7 @@ def __dir__() -> list[str]:
     return __all__
 
 
-BASE_TIMEOUT = 5
+BASE_TIMEOUT = 10
```
henryiii: This doubles all the timeouts. I think we should lower the multipliers, or only apply this for apple silicon native.

src/scikit_build_core/program_search.py
henryiii:
```suggestion
        return BASE_TIMEOUT * 3 if _macos_binary_is_x86(executable) else BASE_TIMEOUT * 2
```

This would be the only native change.

DSeaStar:
Thanks — updated. `BASE_TIMEOUT` is back to 5s, and native Apple Silicon is the only path that doubles (`BASE_TIMEOUT * 2`). CI / Windows / Rosetta multipliers are unchanged.

henryiii (approved):

PR #1543, opened 2026-09-04
fix: drop two-token CMAKE_BUILD_TYPE from CMAKE_ARGS

r3wretrhy:
`CMAKE_ARGS` already drops joined `-DCMAKE_BUILD_TYPE` and `-DCMAKE_INSTALL_PREFIX` because scikit-build-core sets those itself. The two-token form (`-D CMAKE_BUILD_TYPE=Release`) still reached CMake.

Parse it the same way `get_archs` already handles two-token `-D CMAKE_SYSTEM_PROCESSOR`. A prefix match on `-DCMAKE_BUILD_TYPE` also false-positived on `-DCMAKE_BUILD_TYPE_FOO`.

## Test plan
- [x] `uv run pytest tests/test_builder.py::test_builder_get_cmake_args tests/test_builder.py::test_builder_filter_env_cmake_args` (11 passed)
- [x] `uv run ruff check` / `ruff format --check` on the two changed files

### DIFF
--- a/docs/api/scikit_build_core.builder.rst
+++ b/docs/api/scikit_build_core.builder.rst
@@ -17,6 +17,14 @@ scikit\_build\_core.builder.builder module
    :show-inheritance:
    :undoc-members:
 
+scikit\_build\_core.builder.cmake\_args module
+----------------------------------------------
+
+.. automodule:: scikit_build_core.builder.cmake_args
+   :members:
+   :show-inheritance:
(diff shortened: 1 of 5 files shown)


henryiii:
Thanks! I will push a simplification, and also expand this to support CMAKE_OSX_ARCHITECTURES too, same issue.

henryiii (approved):

Issue #1541, opened 2026-09-03
Fable 5.1 xhigh review of scikit-build-core


henryiii:
:robot: _AI text below_ :robot:

Review done: 82 findings across the source tree, 42 of them bugs, 27 reproduced by running a script. Full report with fix sketches and suggested PR grouping is at `/private/tmp/claude-501/-Users-henryfs-git-scikit-build-scikit-build-core/08f0419f-8e1e-45da-82e1-8f417d786c19/scratchpad/review-2026-09-03.md`. Nothing in the repo was modified.

## Bugs

**Settings**

1. `if.implementation-version` / `if.scikit-build-version` never match RC or dev versions on packaging <= 25.0 because `skbuild_overrides.py:98` omits `prereleases=True`. Masked locally by packaging 26.
2. TOML string in a bool field is silently `True` (`experimental = "false"`), `sources.py:776`.
3. `EnvSource(env={})` falls back to `os.environ`, `sources.py:355`. Every other `env=` consumer treats `{}` as empty.
4. `cmake = ">=3.15"` or `wheel = "x"` (scalar where a table is expected) gives a raw traceback, `skbuild_read_settings.py:318`, `sources.py:710`.
5. Unknown key in `[[generate]]` reports "missing 1 required positional argument: 'target'" and is never flagged as unrecognized, `sources.py:735`.
6. `minimum-version = "build-system.requires"` crashes on `scikit-build-core==0.10.*`, `auto_requires.py:58`.
7. `cmake_minimum_required(VERSION ${VAR})` crashes the default settings path, `skbuild_read_settings.py:508`. Two reviewers hit this independently.
8. Trailing `;` in env/config list or dict values gives an empty element or a bare `ValueError`; four copies of the split logic, `sources.py:392,412,573,607`.
9. `inherit.variant` / `inherit.generate` work at runtime but the schema rejects them, `skbuild_schema.py:179`.
10. `_process_bool` and `strtobool` disagree: `SKBUILD_EXPERIMENTAL=n` enables experimental, `sources.py:141` vs `skbuild_overrides.py:81`.
11. `Optional[Union[A, B]]` renders as just `A` in generated docs, so `wheel.packages` is documented as list-only, `documentation.py:89`.
12. MEDIUM/LOW: `cmake.define {env=}` ignores the `env=` argument (`sources.py:276`); repeated `-Ccmake.define.FOO` stringifies a list (`sources.py:543`).

**Build, wheel, sdist, editable**

[...]

PR #1545, opened 2026-09-04
perf: cache entry-point scans and known-wheel lookups

henryiii:

Part of #1541 (items 42 and 43).

`importlib.metadata.entry_points()` scans every installed distribution. It was called once per `SettingsReader` (3-4 per build), three times per CMake configure, and once per dynamic-metadata provider. A new `cached_entry_points` helper in the `_compat` shim does the scan one time per group.

The `if.cmake-wheel` override also re-read `known_wheels.toml` and recomputed `packaging.tags.sys_tags()` for every override. The cached `known_wheels` and `is_known_platform` helpers move from `get_requires` to `builder/_known_wheels.py` (to avoid a circular import), and the override now uses them. Match semantics are unchanged; only the log message no longer lists the matched platform tags.

A conftest fixture clears both caches around each test, because some tests fake the entry points or the system tags.

### DIFF
--- a/src/scikit_build_core/_compat/importlib/metadata.py
+++ b/src/scikit_build_core/_compat/importlib/metadata.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import functools
 import importlib.metadata
 import sys
 
@@ -13,7 +14,7 @@ if TYPE_CHECKING:
     else:
         from importlib.metadata import EntryPoints
 
-__all__ = ["entry_points"]
+__all__ = ["cached_entry_points", "entry_points"]
 
 
 def entry_points(*, group: str) -> EntryPoints:
@@ -22,6 +23,16 @@ def entry_points(*, group: str) -> EntryPoints:
 
     epg = importlib.metadata.entry_points()
     return epg.get(group, [])  # pylint: disable=no-member
+
+
+@functools.cache
+def cached_entry_points(*, group: str) -> EntryPoints:
+    """Like ``entry_points``, but scans the installed distributions only once.
+
+    ``entry_points`` is read from the module global, so a monkeypatch of it is
+    seen. Call ``cached_entry_points.cache_clear()`` if the entry points change.
+    """
+    return entry_points(group=group)
 
 
 def __dir__() -> list[str]:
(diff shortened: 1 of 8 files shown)


LecrisUT:
Initially when reading the description i was thinking that it is weird and if the caching shouldn't be done on a layer higher. Reading the changes though, is there any cache being hit? It seems like in each usage it reads for a different group.

henryiii:
Did a quick investigation, TL;DR is that this does hit since we construct multiple readers. But caching a layer higher would be better. Also, there's some caching of entrypoints, but this still saves a little time.

**Does the cache get hit?** Yes, within one process, but not across the three CMake search groups. Call sites and their groups:

- Settings reader: `scikit_build.overrides` and `scikit_build.default`, once each per `SettingsReader`. A build constructs 3 to 4 readers, so calls 2 onward hit.
- Dynamic metadata: `dynamic_metadata.provider`, once per configured provider. Hits from the second provider on.
- Builder: `cmake.module`, `cmake.prefix`, `cmake.root`. These are three distinct groups, so they never share a cache entry. They only hit when `Builder.configure` runs more than once in a process (tests, editable rebuild).

LecrisUT's "layer higher" idea is valid: the stdlib `entry_points()` already gathers all groups and then filters with `.select(group=...)`, so caching the unfiltered result once and selecting per group would let the three builder groups share one scan. Each hook runs in its own process under a PEP 517 frontend, so the total saving is bounded either way.

**Is there caching in Python already?** Partly. `importlib.metadata.FastPath` is `lru_cache`d per `sys.path` root, and its `lookup` is cached keyed on directory mtime, so directory listings are cached. But every `entry_points()` call still constructs a `PathDistribution` per installed package and re-reads each `entry_points.txt` from disk. Measured in this venv with 38 distributions:

| Call                      | Warm time |
| ------------------------- | --------- |
| `entry_points(group=...)` | 1.08 ms   |
| `list(distributions())`   | 0.16 ms   |
| `cached_entry_points`     | ~0        |

So the stdlib avoids the directory walk but not the metadata file reads. The PR's cache removes those. The saving is about a millisecond per call in a small venv, and it grows with the number of installed packages.

PR #1546, opened 2026-09-04
perf: cut editable shim import time

henryiii:

Part of #1541 (item 40).

Every editable install loads `_editable_redirect.py` at each interpreter start
through its `.pth` file. The module imported `importlib.abc` only to subclass
`MetaPathFinder` (which pulls in `typing` and more), and `subprocess` for the
optional rebuild.

The two finders are now plain classes (a meta path finder only has to supply
`find_spec`), and `importlib.machinery`, `importlib.util` and `subprocess` are
imported in the functions that use them. Only `os` and `sys` stay at module
level. Behaviour is unchanged.

Warm `python -X importtime -c "import shim"` on the shim file alone:

| Python | before | after |
| --- | --- | --- |
| 3.9 | 10.3 ms | 0.5 ms |
| 3.14 | 4.4 ms | 0.8 ms |
| 3.15 | 8.1 ms | 0.8 ms |

A new test parses the shim with `ast` and asserts that the module-level
imports stay `os` and `sys`. The shim was also smoke tested on 3.9, 3.13, 3.14
and 3.15.

### DIFF
--- a/src/scikit_build_core/resources/_editable_redirect.py
+++ b/src/scikit_build_core/resources/_editable_redirect.py
@@ -1,15 +1,16 @@
 from __future__ import annotations
 
-__lazy_modules__ = {"importlib.machinery", "importlib.util", "subprocess"}
-
-import importlib.abc
-import importlib.machinery
-import importlib.util
 import os
-import subprocess
 import sys
 
 # Import as little as possible, since every usage of Python imports this file.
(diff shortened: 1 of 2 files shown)


LecrisUT:
Does upstream have plans to fix it at the importlib.abc level?

henryiii:
This should be fixable (at least in part) with lazy imports upstream, though haven't checked. Probably something that could be contributed.

PR #1554, opened 2026-09-05
fix: program search, build-dir info, and module detection robustness

henryiii:

Fixes items 22, 26, 28, 29 of #1541.

**22 - program search catches too little.** The CMake and Ninja version probes caught `CalledProcessError`, `PermissionError` and `TimeoutExpired`, but a missing or non-executable path raises `FileNotFoundError`. That escaped the probe, so `CMakeNotFoundError` was unreachable for a bad `CMAKE_EXECUTABLE`. Both probes now catch `OSError` and yield a version-less `Program`, which gives the usual `CMakeNotFoundError`.

**26 - macOS architecture probe leaks output.** The `lipo`/`file` calls on Apple Silicon now use `stderr=subprocess.DEVNULL` and `stdin=subprocess.DEVNULL`, so a shell shim cannot print noise or read the build's input.

**28 - corrupt `.skbuild-info.json` blocks all later builds.** Reading the file suppressed only `FileNotFoundError`, so an empty or truncated file from an interrupted build raised on every later build. The read now also handles `OSError` and `ValueError` (which covers `JSONDecodeError`), rejects a non-dict payload, and uses `.get` for the keys. Any of these marks the cache stale, the same path as a source-directory change, so the build directory is wiped and rebuilt.

**29 - namespace directory read as the PyPI package.** `importlib.util.find_spec("cmake")` returns a namespace spec for a bare `cmake/` directory on `sys.path`, which happens with `python -m`, `PYTHONPATH=.` or `backend-path`. That added `cmake>=3.15` to the build requirements for no reason. The check now also requires `spec.origin is not None`; a namespace package has no origin. Same for `ninja`.

Each fix has a regression test, added first and confirmed to fail without the change.

### DIFF
--- a/src/scikit_build_core/builder/get_requires.py
+++ b/src/scikit_build_core/builder/get_requires.py
@@ -120,8 +120,10 @@ class GetRequires:
         cmake_verset = self.settings.cmake.version
 
         # If the module is already installed (via caching the build
-        # environment, for example), we will use that
-        if importlib.util.find_spec("cmake") is not None:
+        # environment, for example), we will use that. A namespace package
+        # (a bare "cmake" directory on sys.path) has no origin and is not it.
+        cmake_spec = importlib.util.find_spec("cmake")
+        if cmake_spec is not None and cmake_spec.origin is not None:
             yield f"cmake{cmake_verset}"
             return
 
@@ -150,8 +152,10 @@ class GetRequires:
         ninja_verset = self.settings.ninja.version
 
         # If the module is already installed (via caching the build
-        # environment, for example), we will use that
-        if importlib.util.find_spec("ninja") is not None:
+        # environment, for example), we will use that. A namespace package
+        # (a bare "ninja" directory on sys.path) has no origin and is not it.
+        ninja_spec = importlib.util.find_spec("ninja")
+        if ninja_spec is not None and ninja_spec.origin is not None:
             yield f"ninja{ninja_verset}"
             return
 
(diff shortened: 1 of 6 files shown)


src/scikit_build_core/builder/get_requires.py
```diff
@@ -150,8 +152,10 @@ def ninja(self) -> Generator[str, None, None]:
         ninja_verset = self.settings.ninja.version
 
         # If the module is already installed (via caching the build
-        # environment, for example), we will use that
-        if importlib.util.find_spec("ninja") is not None:
+        # environment, for example), we will use that. A namespace package
+        # (a bare "ninja" directory on sys.path) has no origin and is not it.
```
henryiii:
```suggestion
        # (a bare "ninja" directory on sys.path) has no origin.
```

src/scikit_build_core/builder/get_requires.py
```diff
@@ -120,8 +120,10 @@ def cmake(self) -> Generator[str, None, None]:
         cmake_verset = self.settings.cmake.version
 
         # If the module is already installed (via caching the build
-        # environment, for example), we will use that
-        if importlib.util.find_spec("cmake") is not None:
+        # environment, for example), we will use that. A namespace package
+        # (a bare "cmake" directory on sys.path) has no origin and is not it.
```
henryiii:
```suggestion
        # (a bare "cmake" directory on sys.path) has no origin.
```

PR #1557, opened 2026-09-05
fix(builder): honor CMAKE_GENERATOR define in requires

henryiii:

Fixes item 23 of #1541. Item 25 (repeated warning) moved to #1573. Item 31 (riscv64 wheel tag) moved to #1563.

The build-requirements code read the generator only from `-G` arguments and
the `CMAKE_GENERATOR` environment variable. The build also reads
`cmake.define.CMAKE_GENERATOR`, so a project that sets the define to Ninja on
Windows got no `ninja` requirement and then failed to configure with
`NinjaNotFoundError`. A new shared `resolve_generator` helper in
`builder/generator.py` applies one precedence for both paths: `-G`, then the
define, then the environment. `CMaker.get_generator` now uses it too, in place
of its own copy of the logic.

### DIFF
--- a/src/scikit_build_core/builder/builder.py
+++ b/src/scikit_build_core/builder/builder.py
@@ -22,6 +22,7 @@ __lazy_modules__ = {
 
 import dataclasses
 import enum
+import functools
 import os
 import platform
 import re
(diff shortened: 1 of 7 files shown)


src/scikit_build_core/builder/builder.py
```diff
@@ -139,6 +140,12 @@ def _warn_macos_arch_mismatch(cmake_path: Path, *, explicit_arch: bool) -> None:
 _UNSUPPORTED_ENV_DEFINES = frozenset({"CMAKE_BUILD_TYPE", "CMAKE_INSTALL_PREFIX"})
 
 
+@functools.cache
+def _warn_unsupported_env_define(ignored: str) -> None:
```
henryiii: Maybe logger.warn_once might make sense? Also I think "default" may warn once? Need to check.

PR #1558, opened 2026-09-05
fix: honor staged Python development environment hints

dajiaohuang:
## Summary

Honor non-empty `PYTHON_INCLUDE_DIR` and `PYTHON_LIBRARY` from the effective build environment when constructing Python development hints. This addresses the staged-path selection described in #952 and the maintainer's proposed direction there.

- Pass the effective environment to include-directory lookup and prefer explicit regular-library hints over automatic detection.
- Pass an explicit regular-library hint to modern FindPython on POSIX as well as Windows, without adding automatic POSIX library hints.
- Preserve empty-value fallback, existing `DIST_EXTRA_CONFIG` handling, stable-ABI library selection, and explicit CMake define precedence.
- Document the environment hints in the manual cross-compilation guide.

## Validation

The new real-CMake configure regression failed on the base commit because the cache selected host development paths instead of distinct staged paths. It passes with this change, using copied matching host headers and a library.

- Builder tests: 123 passed, including real configure, explicit/automatic POSIX hints, regular/stable ABI, empty fallback, and CMake-define precedence.
- CMake configuration/AST tests: 65 passed, 6 skipped.
- Changed Python files: Ruff check and format passed.
- Sphinx nitpicky HTML build: succeeded, 44 source pages.
- `git diff --check`: passed.

Validation ran on Windows. The broader suite run on the initial implementation reported 1039 passed, 25 skipped, 5 xfailed, and 5 failures: four Windows symlink-privilege failures and one missing local pybind11 CMake package. The pybind11 test passed after correcting local discovery; the final POSIX cache condition was subsequently covered by the full Builder suite. Existing mypy diagnostics/toolchain limitations remain documented; a fully green repository-wide run is not claimed.

This verifies staged path selection through real CMake, not a complete Yocto/BitBake or ARM build. No dependency, toolchain, public configuration setting, or CI change is included.

Refs #952.

### DIFF
--- a/docs/guide/crosscompile.md
+++ b/docs/guide/crosscompile.md
@@ -119,6 +119,12 @@ also need to use `wheel.tags` to manually specify the wheel tags to use for the
 file and `cmake.python-hints = false` if the target python should be detected
 using the toolchain file instead.
 
+For a staged Python installation, set `PYTHON_INCLUDE_DIR` and the regular-build
+`PYTHON_LIBRARY` in the effective build environment. These non-empty values are
+used as Python development hints; an unset or empty value keeps the normal
+`sysconfig`/`DIST_EXTRA_CONFIG` fallback. The regular library hint is not used
+for stable-ABI (`abi3`/`abi3t`) builds.
+
 :::{note}
 
 Because most of the logic in [`FindPython`] is gated by the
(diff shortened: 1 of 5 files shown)


henryiii:
The problem is that these are very commonly set, and could be wrong. I think maybe this could be combined with a setting, such as a new string option for `python-hints = "external"` or something like that.

PR #1561, opened 2026-09-06
fix(templates): abi3 templates need CMake 3.26

henryiii:

Part of item 34 of #1541.

The `abi3` and `abi3t` init templates use `Development.SABIModule`, which needs CMake 3.26, but declared 3.15 as the minimum. The declared minimum also selects the CMake to use, so it is now 3.26.

Split out of #1555.


henryiii:
We backport 3.26 unless users disable it, so this does work.

Issue #1565, opened 2026-09-09
Editable redirect: make the package search-location order deterministic, and expose the install tree at runtime


zklaus:
In redirect mode a package that has files in both the source tree and the CMake install tree gets a `__path__` with two entries. The finder builds that list from a `set` ([`ScikitBuildRedirectingFinder.__init__`](https://github.com/scikit-build/scikit-build-core/blob/8872b6860b2873899a29b1e848b4286fb81b1712/src/scikit_build_core/resources/_editable_redirect.py#L450) collects the locations into one, [`find_spec`](https://github.com/scikit-build/scikit-build-core/blob/8872b6860b2873899a29b1e848b4286fb81b1712/src/scikit_build_core/resources/_editable_redirect.py#L487) calls `list()` on it), so the order depends on string hashing, i.e. on `PYTHONHASHSEED`. Minimal reproducer (scikit-build-core 1.0.3, Python 3.10, Linux): a package with one tracked data file that CMake also installs, with different content.

```
pyproject.toml   [build-system] requires = ["scikit-build-core>=1.0"]; build-backend = "scikit_build_core.build"
                 [tool.scikit-build] wheel.packages = ["pkg"]
CMakeLists.txt   file(WRITE "${CMAKE_BINARY_DIR}/both.txt" "from the CMake install tree\n")
                 install(FILES "${CMAKE_BINARY_DIR}/both.txt" DESTINATION "${SKBUILD_PLATLIB_DIR}/pkg/data")
pkg/__init__.py  (empty)
pkg/data/both.txt  "from the source tree"
```

```
$ pip install -e . --no-build-isolation
$ for seed in 0 1 2 3 4 5 6 7; do PYTHONHASHSEED=$seed python -c "
import importlib.util, importlib.resources as r
loc = importlib.util.find_spec('pkg').submodule_search_locations
print(['install' if 'site-packages' in l else 'source' for l in loc], (r.files('pkg')/'data'/'both.txt').read_text().strip())"; done
['source', 'install'] from the source tree
['source', 'install'] from the source tree
['install', 'source'] from the CMake install tree
['source', 'install'] from the source tree
['install', 'source'] from the CMake install tree
['install', 'source'] from the CMake install tree
['install', 'source'] from the CMake install tree
['install', 'source'] from the CMake install tree
```

The same happens with pytorch installed editable (scikit-build-core 1.0.0), where the file that differs between the trees is whatever a stale checkout still holds.

[...]

PR #1567, opened 2026-09-09
feat(editable): expose search paths on redirect loaders

henryiii:

This addresses the second part of #1565. The ordering part is a separate PR.

In editable installs, every loader the finder hands out now has a `paths` attribute. It lists the package's search locations (its `__path__` entries) in `__path__` order: the CMake install tree and the source tree. A plain module has `paths == []`. The attribute is a copy, so a caller cannot change the finder's state. This gives a package a supported way to find the CMake install tree at runtime, without an import of scikit-build-core.

`_SkbuildMultiplexedPath` (the fallback for Python < 3.12) gets a read-only `paths` property with the same shape as python/importlib_resources#310.

The inplace finder already wraps its loaders for `rebuild()`, so its loaders get `paths` too.

Unit tests for the loader kinds live in `tests/test_editable_redirect.py`, next to the existing finder tests. The integration test in `tests/test_editable.py` checks the attribute through a real install.


docs/about/changelog.md
```diff
@@ -1,5 +1,13 @@
 # Changelog
 
+## Version 1.1.0 (in progress)
+
+Features:
+
+- Expose a package's search locations as `module.__loader__.paths` in editable
+  installs, and add `paths` to the pre-3.12 multiplexed resource path by
+  @henryiii in #1565
+
```
henryiii:
```suggestion
```

PR #1568, opened 2026-09-15
test: allow setuptools-scm file_finders deprecation

henryiii:

setuptools-scm 10.3.2 (released today) warns from its global `setuptools.file_finders` entry point on every setuptools `egg_info` run. Because `filterwarnings = error`, this fails `tests/test_setuptools_pep517.py` and `tests/test_setuptools_abi3.py` on every open PR.

The warning tells the project to configure setuptools-scm, but it fires for sample packages in `tests/packages/` that do not use setuptools-scm and pin their own version. A global entry point cannot be silenced from those packages, so this demotes the warning to `default` instead, matching the other setuptools-caused entries.

Verified by running the setuptools tests against setuptools-scm 10.3.2.


henryiii:
I considered moving to vcs_versioning, but since modern version of vcs_versioning support dynamic-versioning directly, that's probably best. The error is harmless since we don't use file_finders.

PR #1570, opened 2026-09-17
fix(variants): sync PEP 817 support with latest updates to variantlib

CavRiley:
Syncs the experimental PEP 817 variant support with its successor PEPs
(825/9999/9998) and the current variantlib API:

- `get_requires_for_build_*` hooks now also return the provider packages from `[variant.providers.*]`, mentioned in PEP 9998.
- Replace the removed `get_variant_label()` / `variant_label=` with the label on `VariantDescription` from the variantlib api.

Draft because  variantlib with this API has no PyPI release yet, so `VARIANTLIB_BUILD_REQUIREMENT` temporarily pins the GitHub tarball


henryiii:
Should we cap the max version of variantlib, probably <0.1, for now, will this be a breaking change?

(Should we have a release first with `VARIANTLIB_BUILD_REQUIREMENT = "variantlib<0.1"`, in other words)

CavRiley:
> Should we cap the max version of variantlib, probably <0.1, for now, will this be a breaking change?
> 
> 
> 
> (Should we have a release with `VARIANTLIB_BUILD_REQUIREMENT = "variantlib<0.1"`, in other words)

I think that is a good idea for the moment as the api evolves. Then once an updated variantlib version is released this may be incorporated.

mgorny:
Thanks. One more thing to consider is to switch to the "possible variants read from `pyproject.toml`" API rather than specifying properties on command line. Them the interface is limited to `variant-label` config setting.

CavRiley:
Thank you for that @mgorny. @henryiii would it be preferable to support both methods of specifying variants?

henryiii:
Do you have a link to the draft for that?

CavRiley:
> Do you have a link to the draft for that?

The copy of the draft PEP is in the wheelnext org [here](https://wheelnext-peps--44.org.readthedocs.build/en/44/pep-9998/#build-backend-support)

I can provide a possible implementation later this week.

henryiii:
> I can provide a possible implementation later this week.

It would probably be good as a followup. I'm still not sure what's in the currently provisionally accepted PEP and what's in a draft PEP; anything that's in the provisionally accepted PEP we could allow (eventually) without the `experimental=true` flag. Anything in a unaccepted PEP does require that flag.

mgorny:
The provisionally accepted PEP only covers some of the installer logic and the `default-priorities` and `variants` parts of metadata. Drafts cover `provider` part, `pyproject.toml` and build backend integration. So you can't really build anything useful of PEP 825 alone.

PR #1571, opened 2026-09-17
fix(variants): cap variantlib build requirement to <0.1

CavRiley:
Caps the injected variant build requirement to `variantlib<0.1`, so released versions keep working with the current variantlib API when the latest updates are published. Split out from the discussion in #1570, which will update the variant support to the new API once released.

Not breaking: this resolves to the same version bare `variantlib` already refers to.


henryiii (approved):

henryiii:
Thanks!

